PR #22, opened 2026-04-24
Fix THINGS-images download by unlinking pre-created temp zip

### LINKED ISSUE #21: [neuralfetch] `download_things_images()` does not download or extract
### Description

`download_things_images()` fails because the temp zip file is created before calling `pooch.retrieve()`. In this setup, `pooch.retrieve()` returns without populating the pre-created file, so extraction then runs on a zero-byte file.

### Steps to reproduce

```python
import os
from pathlib import Path
from neuralfetch.utils import download_things_images

os.environ["NEURALFETCH_THINGS_PASSWORD"] = "<valid password from password_images.txt>"

study_path = Path("/tmp/neuralfetch-repro/study")
study_path.mkdir(parents=True, exist_ok=True)

download_things_images(study_path, fail_on_error=True)


Current behavior: no progress bar for download, fails on extraction.
```

### Environment

- OS: Linux, MacOS
- Python version: 3.12.0
- neuralset version (`pip show neuralset`): 0.1.0
- Relevant package versions:

saarangp:
## What does this PR do?

Fix `download_things_images()` so the temporary zip path is unlinked before
calling `download_file()`.

Without this, the temp file already exists at size 0 when `pooch.retrieve(...)`
is invoked, and the THINGS archive download can result in an empty file that
later fails extraction as an invalid zip.

## Related Issues

Fixes #21 

## Checklist

- [ ] Tests added or updated

### DIFF
--- a/neuralfetch-repo/neuralfetch/utils.py
+++ b/neuralfetch-repo/neuralfetch/utils.py
@@ -415,6 +415,7 @@ def download_things_images(
 
     with tempfile.NamedTemporaryFile(suffix=".zip", delete=False) as tmp_file:
         tmp_path = Path(tmp_file.name)
+    tmp_path.unlink()
 
     try:
         download_file(zip_url, tmp_path, show_progress=True)


teonbrooks (approved):
That makes sense. Thanks for the catch!

kingjr:
Thanks @saarangp !

PR #25, opened 2026-04-29
Improve Datalad downloader for more flexibility.

bpinsard:
## What does this PR do?

- use datalad api rather than subprocess
- allow use of wildcard for files, for flexible subset download (eg. only one output space from fMRIPrep).

## Checklist

- [ ] Tests added or updated

### DIFF
--- a/neuralfetch-repo/neuralfetch/download.py
+++ b/neuralfetch-repo/neuralfetch/download.py
@@ -366,10 +366,6 @@ class BaseDownload(_Module):
         raise NotImplementedError
 
 
-class Wildcard(BaseModel):
-    folder: str
-
-
 class S3(BaseDownload):
     """Download files from an S3 bucket via ``boto3``.
 
@@ -533,8 +529,8 @@ class Datalad(BaseDownload):
         URL of the DataLad repository to clone.
     threads : int
         Number of parallel download threads for ``datalad get`` (default: 1).
-    folders : list of str or Wildcard
-        Directories or glob patterns to retrieve after cloning.  When empty,
+    paths : list of str
+        paths or glob patterns to retrieve after cloning.  When empty,
         all content is retrieved via ``datalad get "*"``.
     """
 
@@ -543,8 +539,8 @@ class Datalad(BaseDownload):
     repo_url: str
     # number of threads used for the datalad operations
     threads: int = 1
-    # list of folders (or wildcards) to actually clone
-    folders: list[str | Wildcard] = []
+    # list of path to actually fetch
+    paths: list[str] = []
 
     @pydantic.computed_field  # type: ignore
     @property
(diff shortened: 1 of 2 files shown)


neuralfetch-repo/neuralfetch/download.py
kingjr, line 408: IIUC we should probably add a datalad requirement here

neuralfetch-repo/neuralfetch/download.py
```diff
@@ -588,30 +567,24 @@ def _download(self) -> None:
...
-        # download
-        for item in tqdm(all_folders, desc=f"Downloading {self.study}", ncols=100):
-            if not item.is_dir():
-                continue
-            self._dl_item(item)
-
-        print("\nDownloaded Dataset")
+        expanded_paths = []
+        # expand paths
+        if len(self.paths) < 1:
```
kingjr, line 574: if self.paths: for path in self.paths ...
else:
expanded_paths

kingjr:
Thanks @bpinsard ! 
@teonbrooks does this look ok to you?

neuralfetch-repo/neuralfetch/download.py
teonbrooks: we can update it at https://github.com/facebookresearch/neuroai/pull/25/changes#diff-0dbbbd82db76c8270a24fa2aa7ac14090bae86119715b3fa669db494f5ecdfaaR540

neuralfetch-repo/neuralfetch/download.py
```diff
@@ -536,8 +532,8 @@ class Datalad(BaseDownload):
         URL of the DataLad repository to clone.
     threads : int
         Number of parallel download threads for ``datalad get`` (default: 1).
-    folders : list of str or Wildcard
-        Directories or glob patterns to retrieve after cloning.  When empty,
+    paths : list of str
```
teonbrooks, line 532: I was actually thinking about a revamp for selective file downloading. we have include with Openneuro which can take a list of patterns. it seems like this might be a useful then to have as a mixin or added to the BaseDownload so that we have a standard way to filter file downloads

bpinsard: Yes, there is likely factorization possible for granular downloads across downloaders that can support it. OpenNeuro and DANDI would support that through Datalad.

teonbrooks: we currently two ways to download from OpenNeuro, (1) openneuro-py to download, which uses pooch underneath the hood and (2) datalad. I think most of our new studies are using (1). 

I think for now the paths approach is fine, but may be refactored later.

teonbrooks:
I created https://github.com/facebookresearch/neuroai/issues/51 as a follow up so it could be tackled after this PR.

I think there are just some minor changes left and the PR will be set to go

teonbrooks:
thanks for your work on this @bpinsard. would you like to take another pass at this? just gave this PR permissions to run the workflow so you should be able to test it against our CI now. also, there's one outstanding comment from JR above related to simplifying the code block

teonbrooks:
Superseded by #258, which folds in these commits (thanks @bpinsard!) and generalizes the `datalad.api` move into a unified `include`/`exclude` selective-download interface across all backends (also fixes the missing-`datalad` silent-success bug). Closing this in favor of #258.

Issue #28, opened 2026-04-30
NeuralFetch: Li2022PetitSample missing language in Word events


kingjr:
### Description

Word events should contain the language, not just Text events

### Steps to reproduce

```python
from pathlib import Path
import neuralset as ns

CACHE = Path.home() / ".cache" / "neuralset"
infra = {"folder": str(CACHE)}
study = ns.Study(name="Li2022PetitSample", path=CACHE)
study.download()
events = study.run()
events.query('type=="Word"').language.unique() == 'english'
```

### Environment

- OS:
- Python version:
- neuralset version (`pip show neuralset`):
- Relevant package versions:

PR #30, opened 2026-04-30
fix: add language to Word events in Li2022Petit study

BelCorentin:
## Problem

Word events loaded from `Li2022Petit` (and `Li2022PetitSample`) had no `language` set. The language string was only attached to the `Text` event, so any downstream code querying `events.query('type=="Word"').language` would get empty strings.

## Fix

Set `events["language"] = language` on the Word DataFrame right after the `lang → language` mapping is resolved, before merging with the other event types.

## Reproducer

```python
events = study.run()
assert events.query('type=="Word"').language.unique().tolist() == ['english']
```


BelCorentin:
Fixes #28

PR #50, opened 2026-05-08
Add emg/qwerty CTC keystroke decoding task

bruAristimunha:
Adds the **emg/qwerty** task: surface-EMG → QWERTY keystroke decoding (CTC, Sivakumar et al., NeurIPS 2024 D&B Track), with auto-fetch via eegdash, four ``--dataset`` recipes (paper-faithful + fine-tune-from-pretrained + compact 51-class vocabulary), and the core-side wiring (Pydantic-discriminated train callbacks, per-extractor CTC metric registry, ``n_outputs_override``, hardened ``load_checkpoint``).

55 unit tests pass.


bruAristimunha:
hey guys (@jrapin, @hubertjb and @kingjr)

Please don't review yet. I will consolidate the results again now using Neural Bench for the competition.

bruAristimunha:
Hey @hubertjb, @jarod1212 and @kingjr!

Now it is ready for the revision! 

@hubertjb, I think you want to run internally to check if it is working as expected! You will need to install the braindecode GitHub version for everything to work. Once you give me the okay, I can release it on the braindecode side.

bruAristimunha:
hey @hubertjb, @jarod1212, @kingjr and @agramfort!

We need to merge this to update the webpage of the proposal :) 

In the proposal, we mentioned that everything is ready in the public datasets.

I was wondering, can we review and merge this week this PR?

docs/neuralbench/tasks/emg/qwerty.rst
```diff
@@ -0,0 +1,71 @@
+Qwerty keystroke decoding
+=========================
+
+| **Name**: qwerty
```
hubertjb: Let's call the task `typing` instead of `qwerty` as we already have an EEG and an MEG task with a similar setup. We'll need to change the docs and filenames.

docs/neuralbench/tasks/emg/qwerty.rst
```diff
@@ -0,0 +1,71 @@
...
+.. code-block:: bash
+
+   # Auto-fetch NM000104 (~239 GB) via eegdash
+   neuralbench emg qwerty -m emg2qwerty --download
+
+   # Local 2-epoch sanity check
+   neuralbench emg qwerty -m emg2qwerty --debug
+
+   # Personalized (N-1 sessions of one subject)
+   neuralbench emg qwerty -m emg2qwerty --dataset paper_personalized
```
hubertjb: I'd be in favour of keeping a single task configuration (apart from debug), at least for the core neuralbench version. Which one are you intending to use ultimately?

docs/neuralbench/tasks/emg/qwerty.rst
```diff
@@ -0,0 +1,71 @@
...
+~~~~~~~~~~~
+
+Continuous-keystroke decoding from 32-channel surface EMG (two 16-electrode
+wristbands at 2 kHz) using the CTC framework from [Sivakumar2024]_.  Each
+5-s window (0.9 s + 4 s core + 0.1 s) is mapped to a variable-length
+keystroke sequence over the 98-key paper vocabulary plus a CTC blank ―
+99 output classes.
+
+The 51-class compact variant case-folds letters, US-QWERTY-folds shifted
+symbols (``!`` → ``1``, ``~`` → `` ` ``, …), and drops ``Key.shift`` ―
```
hubertjb: I think you may need an escape character
```suggestion
symbols (``!`` → ``1``, ``~`` → `` \` ``, …), and drops ``Key.shift`` ―
```

docs/neuralbench/tasks/emg/qwerty.rst
```diff
@@ -0,0 +1,71 @@
...
+
+* **Auto-fetch**: ``--download`` pulls NM000104 from NEMAR
+  (``s3://nemar/nm000104``) via :py:class:`neuralfetch.download.Eegdash`
+  ― 1136 files, ~239 GB, under ``<DATA_DIR>/Emg2qwerty/download/sub-*/…``.
+  An existing BIDS tree placed directly under ``<DATA_DIR>/Emg2qwerty/``
+  is also picked up.
+* **BIDS-aware reader**: the Study reads via
+  :py:func:`mne_bids.read_raw_bids` (``≥ 0.19``); channel types and EMG
+  units come from the BIDS sidecars, no manual coercion or rescaling.
+* **CTC on MPS**: ``nn.CTCLoss`` isn't implemented on Apple MPS.
```
hubertjb: Note to be moved to the CTC loss implementation instead.

docs/neuralbench/tasks/emg/qwerty.rst
```diff
@@ -0,0 +1,71 @@
...
+  ``PYTORCH_ENABLE_MPS_FALLBACK=1`` keeps the rest of the model on MPS
+  while CTC falls back to CPU; paper-level CER still requires a GPU.
+* **Test-time windowing**: the paper trains on 4-s windows but feeds
```
hubertjb: Description above says 5-s windows?

Also, I'd cite the reference again after "paper".

docs/neuralbench/tasks/emg/qwerty.rst
```diff
@@ -0,0 +1,71 @@
...
+  whole sessions at test time.  We apply the same 4-s window across all
+  splits ― slightly pessimistic CER; tracked as a follow-up.
+
+References
+~~~~~~~~~~
+
+.. [Sivakumar2024] Sivakumar et al., "emg2qwerty: A Large Dataset with
```
hubertjb: It's not consistent across the codebase yet, but I'm planning on enforcing MLA style citations:
```
Sivakumar, Viswanath, et al. "emg2qwerty: A large dataset with baselines for touch typing using surface electromyography." Advances in Neural Information Processing Systems 37 (2024): 91373-91389.
```

neuralbench-repo/neuralbench/tasks/emg/typing/config.yaml
```diff
@@ -0,0 +1,90 @@
...
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+# emg2qwerty CTC task: surface-EMG → QWERTY keystroke sequence.
+
+# Paper recipe optimizer (Adam + OneCycleLR, lr 1e-8 → 1e-3 → 1e-6 with
+# 40 % warmup).  Without the slow warmup CTC collapses to all-blank
+# within an epoch.
+lightning_optimizer_config:
```
hubertjb, line 14: We can leave it as is for now since this is the only EMG task so far, but just a note that we'll ultimately want a single preprocessing pipeline for EMG - and so this block should not be defined at the task level.

neuralbench-repo/neuralbench/tasks/emg/typing/config.yaml
```diff
@@ -0,0 +1,90 @@
...
+      valid_split_ratio: 0.2
+      test_split_ratio: 0.2
+      valid_random_state: 33
+      test_random_state: 33
+  neuro:
+    =replace=: true
+    name: EmgExtractor
+    picks: [emg]
+    frequency: 2000.0
+    filter: null
```
hubertjb, line 65: Can you confirm the reference paper didn't use any filtering/scaling/clamping?

neuralbench-repo/neuralbench/tasks/emg/qwerty/test_ctc_pipeline.py
```diff
@@ -0,0 +1,117 @@
...
+    paper_keys = {k for k, _ in PAPER_KEY_TO_LABEL}
+    assert (PAPER_NULL_CLASS, PAPER_NUM_CLASSES, len(paper_keys)) == (98, 99, 98)
+
+    compact_keys = {k for k, _ in COMPACT_KEY_TO_LABEL}
+    assert (COMPACT_NULL_CLASS, COMPACT_NUM_CLASSES, len(compact_keys)) == (50, 51, 50)
+    # compact drops uppercase, shifted symbols, and Key.shift.
+    assert {"Key.shift", "A", "!"}.isdisjoint(compact_keys)
+
+
+# --- Emg2qwerty study (synthetic BIDS tree) ------------------------------
```
hubertjb: These tests below should go in neuralfetch instead as they are data-specific.

neuralbench-repo/neuralbench/tasks/emg/.gitkeep
hubertjb, line 1: To be removed

neuralbench-repo/neuralbench/model_factory.py
```diff
@@ -185,7 +186,9 @@ def build_brain_model(
     LOGGER.info(f"Target shape: {batch.data['target'].shape}")
 
     feat = batch.data["target"]
-    n_outputs = feat.shape[-1]
+    # CTC-style targets carry a length prefix, so ``feat.shape[-1]`` is the vocab
+    # size only for one-hot targets — pass ``n_outputs_override`` for the rest.
+    n_outputs = n_outputs_override if n_outputs_override is not None else feat.shape[-1]
```
hubertjb: I'd like to avoid this override as we should be able to infer the output size from the actual target feature shape.

Can you confirm what shape `feat` has here?

neuralbench-repo/pyproject.toml
```diff
@@ -13,7 +13,10 @@ classifiers = [
 ]
 dependencies = [
   "adjustText",
-  "braindecode",
+  # >=1.5 ships EMG2QwertyNet, exposes braindecode.models.__all__ (needed
```
hubertjb: Good idea to fix this! Let's remove the comment though.

neuralbench-repo/README.md
```diff
@@ -4,7 +4,7 @@
   <img src="assets/hero.png" alt="NeuralBench: open, reproducible benchmarking of NeuroAI models on EEG, MEG, and fMRI -- 36 EEG tasks, 94 EEG datasets (9.5k+ subjects, 13.6k+ hours of recording), and 14 models" width="100%">
 </p>
 
-`neuralbench` is a unified framework to benchmark **NeuroAI models**. It is designed for evaluating pretrained or randomly initialized models on a diverse suite of downstream tasks for brain modeling -- not for pretraining itself. It supports multiple neuroimaging devices -- **EEG**, **MEG**, and **fMRI** -- with more tasks and devices to come.
+`neuralbench` is a unified framework to benchmark **NeuroAI models**. It is designed for evaluating pretrained or randomly initialized models on a diverse suite of downstream tasks for brain modeling -- not for pretraining itself. It supports multiple neuroimaging devices -- **EEG**, **MEG**, **fMRI**, and **EMG** (emg2qwerty CTC keystroke decoding) -- with more tasks and devices to come.
```
hubertjb:
```suggestion
`neuralbench` is a unified framework to benchmark **NeuroAI models**. It is designed for evaluating pretrained or randomly initialized models on a diverse suite of downstream tasks for brain modeling -- not for pretraining itself. It supports multiple neuroimaging devices -- **EEG**, **MEG**, **fMRI**, and **EMG** -- with more tasks and devices to come.
```

neuralfetch-repo/neuralfetch/studies/emg2qwerty.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class Emg2qwerty(study.Study):
+    """emg2qwerty (CTRL-Labs, NeurIPS 2024) — surface-EMG keystroke decoding."""
+
+    url: tp.ClassVar[str] = (
+        "https://proceedings.neurips.cc/paper_files/paper/2024/file/"
+        "a64d53074d011e49af1dfc72c332fe4b-Paper-Datasets_and_Benchmarks_Track.pdf"
+    )
+    bibtex: tp.ClassVar[str] = (
```
hubertjb: For consistency with other studies, let's use triple quotes instead.

neuralfetch-repo/neuralfetch/studies/emg2qwerty.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _read(self) -> tp.Any:
+        bp = mne_bids.get_bids_path_from_fname(self.filepath)
+        return mne_bids.read_raw_bids(bp, verbose=False)
+
+
+class Emg2qwerty(study.Study):
+    """emg2qwerty (CTRL-Labs, NeurIPS 2024) — surface-EMG keystroke decoding."""
+
+    url: tp.ClassVar[str] = (
```
hubertjb: To be moved to bibtex url field instead

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
```diff
@@ -0,0 +1,160 @@
...
+# then fail informatively at use site).
+mne_bids = _soft_import(
+    "mne_bids",
+    "reading the BIDS-formatted NM000104 EMG recordings",
+    strict=False,
+    min_version=_MNE_BIDS_MIN_VERSION,
+)
+
+
+class Emg2qwertyRaw(etypes.Emg):
```
hubertjb: Not sure it's clear from the docs yet but the NeuralFetch convention is to use `<Last name><year><first word of title>` as study class name. Here this should be `Sivakumar2024Emg2qwerty`, and the script should be called `sivakumar2024emg2qwerty.py`.

neuralfetch-repo/neuralfetch/studies/emg2qwerty.py
```diff
@@ -0,0 +1,160 @@
...
+        "108 subjects doing surface typing with an EMG wristband on each arm."
+    )
+    aliases: tp.ClassVar[tuple[str, ...]] = ("emg2qwerty", "nm000104")
+
+    NEMAR_DATASET_ID: tp.ClassVar[str] = "nm000104"
+
+    _bids_root_cache: Path | None = pydantic.PrivateAttr(default=None)
+
+    def _download(self) -> None:
+        from neuralfetch import download
```
hubertjb: Move import to top of file

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
```diff
@@ -0,0 +1,160 @@
...
+    NEMAR_DATASET_ID: tp.ClassVar[str] = "nm000104"
+
+    _bids_root_cache: Path | None = pydantic.PrivateAttr(default=None)
+
+    def _download(self) -> None:
+        from neuralfetch import download
+
+        download.Eegdash(study=self.NEMAR_DATASET_ID, dset_dir=self.path).download()
+
+    def _bids_root(self) -> Path:
```
hubertjb: I'd favour enforcing a single path to simplify (i.e. the one obtained with `Study.download`). Users with an independently downloaded directory should rather create a symlink if necessary.

neuralfetch-repo/neuralfetch/studies/emg2qwerty.py
```diff
@@ -0,0 +1,160 @@
...
+            "language": "en",
+        })
+
+        raw_row = pd.DataFrame([{
+            "type": "Emg2qwertyRaw", "filepath": str(bp.fpath),
+            "start": 0.0, "subject": timeline["subject"],
+        }])
+        return (
+            pd.concat([raw_row, sentences, keystrokes], ignore_index=True)
+            .sort_values("start").reset_index(drop=True)
```
hubertjb: The sorting is done by default by the parent class so we can remove it here.

neuraltrain-repo/neuraltrain/augmentations/augmentations.py
```diff
@@ -169,6 +169,78 @@ def forward(self, x: torch.Tensor):
...
+        self.electrodes_per_band = electrodes_per_band
+        self.band_offsets = tuple(band_offsets)
+        self.max_temporal_jitter = max_temporal_jitter
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if not self.training:
+            return x
+        if x.shape[1] != self.num_bands * self.electrodes_per_band:
+            return x
+        from braindecode.augmentation.functional import band_rotation
```
hubertjb: Should be imported in init instead

neuraltrain-repo/neuraltrain/augmentations/augmentations.py
```diff
@@ -169,6 +169,78 @@ def forward(self, x: torch.Tensor):
...
+    def build(self) -> nn.Module:
+        return BandRotation(
+            num_bands=self.num_bands,
+            electrodes_per_band=self.electrodes_per_band,
+            band_offsets=self.band_offsets,
+            max_temporal_jitter=self.max_temporal_jitter,
+        )
+
+
+class BandRotation(nn.Module):
```
hubertjb: Any chance band_rotation could be implemented as a class in braindecode? This way we could follow the same pattern as other augmentations and just define a config here.

neuraltrain-repo/neuraltrain/augmentations/augmentations.py
```diff
@@ -169,6 +169,78 @@ def forward(self, x: torch.Tensor):
         return iffted
 
 
+class BandRotationConfig(pydantic.BaseModel):
```
hubertjb: Is this augmentation necessary to get reasonable baseline performance? If possible, I'd favour taking it out of the baseline task config.

neuraltrain-repo/neuraltrain/metrics/metrics.py
```diff
@@ -730,3 +730,56 @@ def reset(self) -> None:
 
     def __repr__(self) -> str:
         return f"GroupedMetric({self.base_metric_cls.__name__})"
+
+
+class CharacterErrorRates(torchmetrics.Metric):
```
hubertjb: Can we get the same result with torchmetrics's https://github.com/facebookresearch/neuroai/pull/50/changes#diff-39a20ffaa530d256d8f5b390b26bbb025c266be02463aba8ad6a309e0cf76f75 ?

neuraltrain-repo/neuraltrain/utils.py
```diff
@@ -86,6 +86,14 @@ def build_method(instance: BaseModel):
             for field in type(instance).model_fields
             if (field != "name" and field not in exclude_from_build)
         )
+        # Wrapper-only fields that must not reach the underlying class.
```
hubertjb: Can you explain with an example why we need this?

hubertjb (commented):
Thanks @bruAristimunha ! I did a first pass and left a couple of comments if you can take a look. I've asked @jarod1212 to look into some specifics next - will get back to you soon on this.

neuraltrain-repo/neuraltrain/utils.py
bruAristimunha: It was mostly because of the data augmentation, but I am double-checking if we really need.

neuraltrain-repo/neuraltrain/metrics/metrics.py
bruAristimunha: we can!

neuraltrain-repo/neuraltrain/augmentations/augmentations.py
bruAristimunha: removed!

neuraltrain-repo/neuraltrain/augmentations/augmentations.py
bruAristimunha: i implemented!

neuralbench-repo/pyproject.toml
```diff
@@ -13,7 +13,7 @@ classifiers = [
 ]
 dependencies = [
   "adjustText",
-  "braindecode",
+  "braindecode>=1.5.0",
```
bruAristimunha:
```suggestion
  "braindecode>1.5.0",
```

neuralbench-repo/neuralbench/tasks/emg/typing/charset.py
```diff
@@ -0,0 +1,107 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
jarod1212: I think this file is required for future works on this dataset to be comparable with the original paper but it can be much more compact. Notably the qwerty_compact preset is not referenced anywhere in the PR. I think this collapses to ~25 lines. 

Concretely:

```
from __future__ import annotations
import string
from neuralset.extractors.text import Folds, Vocab
_VOCAB: tuple[str, ...] = (
    *string.ascii_letters, *string.digits, *string.punctuation,
    "Key.backspace", "Key.enter", "Key.space", "Key.shift",
)
_UNICHAR_ALIASES: dict[str, str | None] = {
    " ": "Key.space", "\n": "Key.enter", "\r": "Key.enter", "\b": "Key.backspace",
    "⇧": "Key.shift", "⏎": "Key.enter", "⌫": "Key.backspace",
}
PAPER_KEY_TO_LABEL: Vocab = list(zip(_VOCAB, range(len(_VOCAB))))
PAPER_UNICHAR_TO_KEY: Folds = list(_UNICHAR_ALIASES.items())
PAPER_NULL_CLASS = len(_VOCAB)        # 98
PAPER_NUM_CLASSES = PAPER_NULL_CLASS + 1    # 99

```

Also a suggestion: renaming it vocab.py is maybe clearer.

neuraltrain-repo/neuraltrain/losses/losses.py
```diff
@@ -276,3 +276,49 @@ def forward(
             loss_values["total"] += self.weights[name] * loss_values[name]
 
         return loss_values
+
+
+class CtcSeqLoss(nn.Module):
```
jarod1212: I think we should rely on the torch.nn.CTCLoss instead of defining a new loss in neuraltrain. 
nn.CTCLoss requires (log_probs, targets, input_lengths, target_lengths) while BrainModule._step requires 2 args. So we need to reorg a bit. I would:

1. Drop CTCSeqLoss and its test.
2. Drop the length-prefix from KeystrokeSequence: output shape goes from (max_target_length +1,) to (max_target_length,) padded with pad_value (which keeps defaulting to len(key_to_label), the blank).
3. In pl_module.py add a CTC branch in BrainModule._run_step alongside the existing `isinstance(self.loss,` nn.CrossEntropyLoss)` like this

```
if isinstance(self.loss, nn.CTCLoss):
    blank = self.loss.blank
    target_lengths = (y_true != blank).sum(dim=-1)
    input_lengths = torch.full(
        (y_true.shape[0],), y_pred.shape[1],
        dtype=torch.long, device=y_pred.device,
    )
    # y_pred is (B, T_out, C) log-probs from the CTC head; CTCLoss wants (T_out, B, C).
    loss = self.loss(y_pred.transpose(0, 1), y_true, input_lengths, target_lengths)
else:
    if isinstance(self.loss, nn.CrossEntropyLoss):
        ...  # existing branch
    elif isinstance(self.loss, nn.BCEWithLogitsLoss):
        ...  # existing branch
    loss = self.loss(y_pred, y_true)
```

4. In CharacterErrorRates.update, need to replace 
```
target_lengths = y_true[:, 0].long().tolist()
y_true[i, 1 : 1 + target_len]
```
with the same: `(y_true != self._blank).sum(-1)`

5. In the config then:
```
loss:
  name: CTCLoss
  kwargs:
    blank: 98
    zero_infinity: true
```

Weakness of the proposal: this works only when the extractor's pad value is equal to the loss blank. Maybe we should drop the pad_value parameter entirely and always use len(key_to_label)

This is a big change. Probably need to be validated with @hubertjb but it avoids the magic `y_true[:, 0]`

neuralset-repo/neuralset/extractors/text.py
```diff
@@ -511,3 +515,169 @@ def part_reversal(tensor: torch.Tensor) -> None:
...
+# Vocab fields accept either list-of-pairs OR dict.  List-of-pairs is the
+# YAML form (it dodges ``exca.ConfDict``'s dot-key-flattening on entries
+# like ``"Key.backspace"`` / ``"."``).  Dict is the natural Python form
+# for tests + library users.  ``model_post_init`` normalizes both into
+# the private ``self._*`` dicts used by :meth:`_encode`.
+Vocab = list[tuple[str, int]] | dict[str, int]
+Folds = list[tuple[str, str | None]] | dict[str, str | None]
+
+
+class KeystrokeSequence(BaseStatic):
```
jarod1212: This feature can be handled via LabelEncoder + aggregation='cat' + fixed-length padding. 

LabelEncoder already handles the encoding (via predefined_mapping) and the concatenation (via aggregation='cat'). 

The only missing piece is the fixed-length output padding which is needed to collate the batch (it uses torch.cat(axis=0)). 

2 proposals: big change on LabelEncoder itself or a new feature much more compact that inherits from LabelEncoder. 

Proposal: add 2 small fields on LabelEncoder itself

```
class LabelEncoder(EventField):
    # ... existing fields
    max_length: int | None = None
    pad_value: int | None = None             # defaults to len(predefined_mapping)
    allow_unknown_labels: bool = False       # OOV → drop+warn instead of raise
    target_window_offset: float = 0.0
    target_window_duration: float | None = None
```

Resulting YAML:

```
target:
  =replace=: true
  name: LabelEncoder
  event_types: Keystroke
  event_field: text
  aggregation: cat
  allow_missing: true
  allow_unknown_labels: true
  max_length: 128
  target_window_offset: 0.9
  target_window_duration: 4.0
  predefined_mapping: !!python/name:neuralbench.tasks.emg.typing.charset.PAPER_KEY_TO_LABEL
```

Also a big change. I am in favor of a new feature for this specific project that inherits from LabelEncoder to have something more compact than the current KeystrokeSequence (currently 170 lines of code).  What do you think @hubertjb ?

neuralbench-repo/neuralbench/models/emg2qwerty.yaml
```diff
@@ -0,0 +1,13 @@
...
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+# EMG2QwertyNet (Sivakumar et al., NeurIPS 2024) — pairs with emg/typing.
+brain_model_config:
+  =replace=: true
+  name: EMG2QwertyNet
+  kwargs:
+    log_softmax: true
+    spec_augment: true
```
jarod1212, line 13: The paper is also using temporal jitter and random band rotation. This will likely under-perform the paper's reported CER if not included ?

jarod1212 (commented):
Hello @bruAristimunha ! Solid work thank you! 
My comments are mostly about avoiding new abstractions in neuraltrain and neuralset.

neuralbench-repo/neuralbench/models/emg2qwerty.yaml
bruAristimunha: We implement this at braindecode, like the band rotation got incorporated inside the model, and the temporal jitter is one of the possible augmentation avaliable at braindecode.

neuralset-repo/neuralset/extractors/text.py
bruAristimunha: I like this proposal! I think it makes much more sense to make the LabelEncoder more robust than the classes I'm currently pushing.

neuraltrain-repo/neuraltrain/losses/losses.py
bruAristimunha: I think it works almost perfectly!

In the original code from @agramfort, they use the nn.CTCLoss.
https://github.com/facebookresearch/emg2qwerty/blob/3200d91eeb952cbed1f278e47d0cc56928334fd1/emg2qwerty/lightning.py#L218

neuralbench-repo/neuralbench/tasks/emg/typing/charset.py
bruAristimunha: The compact was the only way I could find with a colleague who had experience using EMG to make the model work with a few subjects. EMG is more data-hungry than EEG in my tests.

neuralbench-repo/neuralbench/models/emg2qwerty.yaml
bruAristimunha: I tested only with a random band and the other augmentation in the model, and this yielded the result I submitted.

neuralset-repo/neuralset/extractors/text.py
bruAristimunha: working on this rn

docs/neuralbench/tasks/emg/qwerty.rst
bruAristimunha: the source code is 4s

docs/neuralbench/tasks/emg/qwerty.rst
bruAristimunha: could be the original paper only

neuralbench-repo/neuralbench/tasks/emg/typing/config.yaml
bruAristimunha: no filtering.

docs/neuralbench/tasks/emg/qwerty.rst
bruAristimunha: i removed the compact

neuralset-repo/neuralset/extractors/text.py
hubertjb: Thanks for the suggestion @jarod1212 ! I'd be in favour of creating a new feature that inherits from LabelEncoder and that lives in neuralbench only for now (e.g. neuralbench/utils.py). Once this is validated and we see it required in another project we could then move it to neuralset.

docs/neuralbench/tasks/emg/typing.rst
```diff
@@ -0,0 +1,63 @@
...
+Description
+~~~~~~~~~~~
+
+Continuous-keystroke decoding from 32-channel surface EMG (two 16-electrode
+wristbands at 2 kHz) using the CTC framework from [Sivakumar2024]_.  Each
+5-s window (0.9 s + 4 s core + 0.1 s) is mapped to a variable-length
+keystroke sequence over the 98-key paper vocabulary plus a CTC blank ―
+99 output classes.
+
+The benchmark variant restricts training to the first ~10 subjects
```
hubertjb:
```suggestion
As compared to the original paper, the NeuralBench default configuration restricts training to the first ~10 subjects
```

docs/neuralbench/tasks/emg/typing.rst
```diff
@@ -0,0 +1,63 @@
...
+  (``s3://nemar/nm000104``) via :py:class:`neuralfetch.download.Eegdash`
+  ― 1136 files, ~239 GB, under
+  ``<DATA_DIR>/Sivakumar2024Emg2qwerty/download/sub-*/…``.
+  Users with an existing BIDS copy should symlink it into ``download/``.
+* **BIDS-aware reader**: the Study reads via
+  :py:func:`mne_bids.read_raw_bids` (``≥ 0.19``); channel types and EMG
+  units come from the BIDS sidecars, no manual coercion or rescaling.
+* **Test-time windowing**: the paper [Sivakumar2024]_ trains on 4-s
+  windows but feeds whole sessions at test time.  We apply the same
+  4-s core window (inside a 5-s padded crop, 0.9 s + 4 s + 0.1 s)
```
hubertjb: I'm not sure I understand the distinction. The config.yaml is using 5-s windows, should we just describe it as 5-s sliding windows with a 1-s stride?

neuralbench-repo/neuralbench/tasks/emg/typing/vocab.py
hubertjb, line 1: I think we should move this to the study script directly, along with the unit test.

hubertjb: Actually, to simplify even further (and since this comes from the original paper): we could include a "label" column inside the Keystroke event when creating the events dataframe.

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
```diff
@@ -0,0 +1,161 @@
...
+
+import pandas as pd
+import pydantic
+from mne.utils import _soft_import
+
+from neuralfetch import download
+from neuralset.events import etypes, study
+
+LOGGER = logging.getLogger(__name__)
+_MNE_BIDS_MIN_VERSION = "0.19"
```
hubertjb: Instead of this logic and the _soft_import call, I'd be in favour of bumping up the mne-bids version in the pyproject.toml, especially if it's supposed to be faster now.

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
```diff
@@ -0,0 +1,161 @@
...
+    aliases: tp.ClassVar[tuple[str, ...]] = ("emg2qwerty", "nm000104")
+
+    NEMAR_DATASET_ID: tp.ClassVar[str] = "nm000104"
+
+    _bids_root_cache: Path | None = pydantic.PrivateAttr(default=None)
+
+    def _download(self) -> None:
+        download.Eegdash(study=self.NEMAR_DATASET_ID, dset_dir=self.path).download()
+
+    def _bids_root(self) -> Path:
```
hubertjb: A `@property` pattern might be slightly cleaner

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
```diff
@@ -0,0 +1,161 @@
...
+                "text": text[sent_mask],
+                "language": "en",
+            }
+        )
+
+        key = ev["key"].astype("string").str.strip()
+        ks_mask = ev["value"].str.startswith("keystroke_", na=False) & (key != "")
+        keystrokes = pd.DataFrame(
+            {
+                "type": "Keystroke",
```
hubertjb, line 164: To follow up on the comment above - I'd add the "label" as an extra field here.

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
```diff
@@ -0,0 +1,161 @@
...
+# then fail informatively at use site).
+mne_bids = _soft_import(
+    "mne_bids",
+    "reading the BIDS-formatted NM000104 EMG recordings",
+    strict=False,
+    min_version=_MNE_BIDS_MIN_VERSION,
+)
+
+
+class Emg2qwertyRaw(etypes.Emg):
```
hubertjb: What about renaming to `BidsEmg` to make the intention clearer?

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -728,6 +729,13 @@ def get_static(self, event: etypes.Event) -> torch.Tensor:
 class LabelEncoder(EventField):
     """Encode a given field from an event, e.g. to be used as a label.
 
+    With ``aggregation='cat'`` plus ``max_length``, doubles as a fixed-length
```
hubertjb: As mentioned above, I'd be in favour of keeping the CTC-specific logic separate from LabelEncoder, and move it to a new extractor in neuralbench itself for now.

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -748,15 +756,62 @@ class LabelEncoder(EventField):
...
+        → ``len(_label_to_ind)`` (only populated post-``prepare()``).
+        """
+        if self.pad_value is not None:
+            return self.pad_value
+        if self.predefined_mapping is not None:
+            return len(self.predefined_mapping)
+        return len(self._label_to_ind) or None
+
+    @property
+    def num_classes(self) -> int:
```
hubertjb:
```suggestion
    def n_classes(self) -> int:
```

neuralset-repo/neuralset/extractors/meta.py
```diff
@@ -392,6 +392,17 @@ def model_post_init(self, log__: tp.Any) -> None:
         self.frequency = self.extractor.frequency  # type: ignore
         super().model_post_init(log__)
 
+    @property
+    def num_classes(self) -> int:
```
hubertjb:
```suggestion
    def n_classes(self) -> int:
```

neuralset-repo/neuralset/utils.py
```diff
@@ -45,16 +45,6 @@ def match_list(A, B, on_replace="delete"):
     """
     from rapidfuzz.distance.Levenshtein import editops
```
hubertjb, line 47: Why has this been deleted?

neuralset-repo/neuralset/extractors/text.py
```diff
@@ -511,3 +511,5 @@ def part_reversal(tensor: torch.Tensor) -> None:
         x[first : int(last)] = reversed(x[first : int(last)])  # type: ignore
         last += num
         first = int(last)
+
+
```
hubertjb: A couple of files have 2 blank lines at their end

neuraltrain-repo/neuraltrain/utils.py
```diff
@@ -86,6 +86,35 @@ def build_method(instance: BaseModel):
             for field in type(instance).model_fields
             if (field != "name" and field not in exclude_from_build)
         )
+        # Drop wrapper-only metric labels that must not reach the underlying class
```
hubertjb: Not sure why this is necessary - an example would be useful!

hubertjb (commented):
A couple of additional comments, but I think we're almost there. Thanks @bruAristimunha !

docs/neuralbench/tasks/emg/typing.rst
bruAristimunha: could work! I just defined thinking in terms of the n_times dots.

neuralbench-repo/neuralbench/tasks/emg/typing/vocab.py
bruAristimunha: good point :)

neuraltrain-repo/neuraltrain/utils.py
bruAristimunha: Mostly because of some parsing limitation, I will fix it at NeuralBench. Thanks for pointing again.

neuralset-repo/neuralset/utils.py
bruAristimunha: I was playing with this distance to use what exists, but it is not possible.

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
bruAristimunha: i agree!

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
bruAristimunha: done

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
bruAristimunha: done!

neuralset-repo/neuralset/extractors/base.py
bruAristimunha: convertion made :)

neuralfetch-repo/neuralfetch/download.py
```diff
@@ -832,68 +834,19 @@ class Eegdash(BaseDownload):
...
-                dset_dir=Path(self.dset_dir),
-                bucket=bucket,
-                files_with_destinations=file_pairs,
-                anonymous=True,
-                skip_existing=True,
-            )
-            dl._download()
-
-        print(f"\nDownloaded {total_files} files for {self.study}")
+        from eegdash import EEGDashDataset  # type: ignore[import-not-found]
```
hubertjb, line 841: @kingjr fyi this simplifies the downloader substantially

hubertjb:
For reference, I'm getting these results for a full run over 3 seeds, confirmed by @bruAristimunha:

| Model | CER (%) ↓ |
|---|---|
| EMG2QwertyNet | 85.88 ± 2.43 |
| Dummy | 93.05 ± 0.00 |
| Chance | 98.83 ± 2.17 |

hubertjb:
Thanks @bruAristimunha !

PR #57, opened 2026-05-11
Update cache clearing logic in study.py

julien-gadonneix:
## What does this PR do?

Change in the API with `exca >= 0.5.24`

## Checklist

- [x] Tests added or updated

### DIFF
--- a/neuralset-repo/neuralset/events/study.py
+++ b/neuralset-repo/neuralset/events/study.py
@@ -396,8 +396,8 @@ class Study(base.Step):
             raise RuntimeError(f"Directory is empty: {self.path}")
         logger.info(f"Success: Study downloaded to {self.path}.")
         _set_dir_permissions(self.path)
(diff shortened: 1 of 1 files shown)


neuralset-repo/neuralset/events/study.py
```diff
@@ -396,7 +396,10 @@ def download(self, **kwargs: tp.Any) -> None:
             raise RuntimeError(f"Directory is empty: {self.path}")
         logger.info(f"Success: Study downloaded to {self.path}.")
         _set_dir_permissions(self.path)
-        self.clear_cache()
+        if hasattr(base.Step, "query"):  # exca >=0.5.24
+            base.Step.query(self).clear_cache()  # type: ignore[attr-defined]
```
jrapin:
```suggestion
        if hasattr(base.Step, "lookup"):  # exca >=0.5.24
            self.lookup().clear_cache()  # type: ignore[attr-defined]
```

PR #62, opened 2026-05-12
Add CAR (common-average reference) to IeegExtractor

### LINKED ISSUE #61: Add CAR (common-average reference) modes to IeegExtractor
### Description

`IeegExtractor` supports `reference="bipolar"` and exposes `notch_filter / filter / apply_hilbert / scaler / clamp` as kwargs, but common-average reference (global or within-shaft) isn't reachable without a subclass. Would you take a PR extending the existing `Literal`, mirroring the `"bipolar"` pattern?

```python
reference: Literal["bipolar", "car", "car_shaft"] | None = None
```

- `"car"`: subtract across-channel mean (any iEEG picks).
- `"car_shaft"`: subtract within-shaft mean; shaft identity by digit-strip on channel names, same convention as `_apply_bipolar_ref`. Requires `picks=("seeg",)`.

Validation parallels bipolar: `car_shaft` + non-sEEG → `ValueError`; `car` / `car_shaft` + `bipolar_ref` → `ValueError`.

Use case: cross-patient sEEG speech-decoding ablation across reference schemes (raw / bipolar / shaft-CAR / global-CAR) via config swap.

Scope is small (~5 tests, one method, one helper, one branch in `_preprocess_raw`, default unchanged). Prototype branch with tests + ruff clean is ready. Happy to open a PR, or adjust the API first (e.g. separate `car_mode` field, `raw.apply_function` instead of direct data mutation).

### Alternatives considered

- `mne.set_eeg_reference("average")`: doesn't generalize to within-shaft grouping and is EEG-typed-channel only.
- Separate `car_mode` field instead of extending `reference`. Cleaner orthogonality but creates two ways to specify "no reference"; mutual exclusion still needed. Happy to take this shape.
- Local `CARIeegExtractor` subclass. Works, but the `"bipolar"` precedent suggests CAR belongs on `IeegExtractor` too.

bentang18:
## What does this PR do?

Adds global common-average reference (`reference="car"`) to `IeegExtractor`, mirroring the existing `"bipolar"` pattern. Per #61 review, shaft-CAR was dropped — shaft naming has no universal convention, so that belongs in `mne.Projection` user-side, not in NeuralSet.

Diff is +56/−2: one `Literal` extension, one validation clause (`car` + `bipolar_ref` → `ValueError`), one `elif` branch in `_preprocess_raw`, one 18-line `_apply_car_ref` method, two tests.

## Related Issues

Closes #61.

## Checklist

- [x] Tests added or updated

### DIFF
--- a/neuralset-repo/neuralset/extractors/neuro.py
+++ b/neuralset-repo/neuralset/extractors/neuro.py
@@ -718,8 +718,13 @@ class IeegExtractor(MneRaw):
         # mne.set_eeg_reference with ch_type=list-of-str computes one reference
         # from the union of channel types (and per-type independence requires
         # projection=True). Loop per type so each channel type gets its own
-        # average reference.
+        # average reference. Skip types absent from raw (picks may name types
+        # not present after _pick_channels, e.g. picks=("seeg","ecog") with a
+        # seeg-only recording).
+        present_types = set(raw.get_channel_types())
         for ch_type in self.picks:
+            if ch_type not in present_types:
+                continue
             raw, _ = mne.set_eeg_reference(
                 raw,
                 ref_channels="average",
--- a/neuralset-repo/neuralset/extractors/neuro.py
+++ b/neuralset-repo/neuralset/extractors/neuro.py
@@ -174,14 +174,15 @@ class MneRaw(BaseExtractor):
     1. Channel selection
     2. Drop bad channels
     3. Bipolar referencing
-    4. Notch filtering
-    5. Band-pass filtering
-    6. Hilbert transform
-    7. Resampling
-    8. Scaling
-    9. Applying projectors
-    10. Baseline correction (applied on segments)
-    11. Clamp (applied on segments)
+    4. Common-average referencing
+    5. Notch filtering
+    6. Band-pass filtering
+    7. Hilbert transform
+    8. Resampling
+    9. Scaling
+    10. Applying projectors
+    11. Baseline correction (applied on segments)
+    12. Clamp (applied on segments)
 
     Parameters
     ----------
@@ -234,6 +235,13 @@ class MneRaw(BaseExtractor):
         Applied after channel selection and dropping bad channels but before
         filtering. The original monopolar channels consumed by the pairs are
         removed and replaced with the new bipolar channels.
+    car_ref : bool, default=False
+        If True, applies a common-average reference (CAR) via
+        ``mne.set_eeg_reference``. Each CAR-eligible channel type present
+        in the raw (``eeg``, ``ecog``, ``seeg``, ``dbs``) is referenced
+        independently to its own average. Applied after ``bipolar_ref``;
+        the two are mutually exclusive. Not supported on
+        ``MegExtractor``, ``EmgExtractor``, or ``FnirsExtractor``.
     channel_order: ["unique", "original"]
         `if self.channel_order=="original"`
         Assigns channel indices for each raw file (doesn't match channel names across files).
@@ -272,8 +280,16 @@ class MneRaw(BaseExtractor):
     clamp: float | None = None
     fill_non_finite: float | None = None
     bipolar_ref: tuple[list[str], list[str]] | None = None
+    car_ref: bool = False
     channel_order: tp.Literal["unique", "original"] = "unique"
     allow_maxshield: bool = False
+
+    _CAR_ELIGIBLE_CH_TYPES: tp.ClassVar[tuple[str, ...]] = (
+        "eeg",
+        "ecog",
+        "seeg",
+        "dbs",
+    )
 
     _channels: dict[str, int] = {}
 
@@ -303,6 +319,10 @@ class MneRaw(BaseExtractor):
                     f"bipolar_ref anodes and cathodes must have equal length, "
                     f"got {len(anodes)} and {len(cathodes)}"
                 )
+        if self.car_ref and self.bipolar_ref is not None:
+            raise ValueError(
+                "car_ref and bipolar_ref are mutually exclusive; choose one."
+            )
 
     def prepare(self, obj: DataframeOrEventsOrSegments) -> None:
         """Specify how to load and preprocess the event.
@@ -350,6 +370,10 @@ class MneRaw(BaseExtractor):
             raw.load_data()
             anodes, cathodes = self.bipolar_ref
             raw = mne.set_bipolar_reference(raw, anodes, cathodes, verbose="WARNING")
+
+        if self.car_ref:
+            raw.load_data()
+            raw = self._apply_car_ref(raw)
 
         if self.notch_filter is not None:
             raw.load_data()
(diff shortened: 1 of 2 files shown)


neuralset-repo/neuralset/extractors/neuro.py
```diff
@@ -688,6 +697,25 @@ def _apply_bipolar_ref(self, raw: mne.io.Raw) -> mne.io.Raw:
...
+            Raw instance that will be referenced.
+
+        Returns
+        -------
+        raw : mne.io.Raw
+            Referenced Raw object
+
+        """
+        logger.info("Applying CAR reference")
+        raw._data -= raw._data.mean(axis=0, keepdims=True)
```
agramfort: I agree with @kingjr. Don't do this manually. Use an mne function for this that will be smarter eg dealing with channel types.

agramfort (changes requested):

neuralset-repo/neuralset/extractors/neuro.py
bentang18: Done in bbb23f0. Switched to mne.set_eeg_reference(ref_channels="average", ch_type=list(self.picks)), so with multiple picks each channel type is referenced to its own average. Updated the docstring to match.

bentang18: Quick follow-up. The multi-pick test I added caught that with `projection=False`, `set_eeg_reference(ch_type=list)` references against the union of channel types rather than per-type (per-type needs `projection=True`, versionchanged 1.2). Fixed in 8db2709 by looping per type. A second pass caught a regression in that fix where absent channel types in `picks` would error out. Fixed in 4432e37 with a guard test.

julien-gadonneix:
Maybe this step should be included in the general `MneRaw` extractor under `_preprocess_raw`?

bentang18:
Fair point, CAR is general enough that it belongs in the base alongside `bipolar_ref`. Will lift it to `MneRaw` in this PR: `car_ref: bool` on the base, modality guards on Meg/Emg/Fnirs, `IeegExtractor.reference="car"` becomes a thin shortcut over it.

agramfort (approved):

bentang18:
@julien-gadonneix I lifted CAR to the `MneRaw` base as you suggested, in commit `3079337`: `car_ref` on the base extractor, modality guards on Meg/Emg/Fnirs, and `IeegExtractor.reference="car"` is now a thin shortcut over it. agramfort has re-approved the result. Let me know if the shape looks right to you.

neuralset-repo/neuralset/extractors/neuro.py
```diff
@@ -585,9 +664,12 @@ class IeegExtractor(MneRaw):
     ----------
     picks: default = ("seeg", "ecog", )
         pick "seeg" and "ecog" channels by default.
-    reference: "bipolar" or None, default=None
+    reference: "bipolar", "car", or None, default=None
```
julien-gadonneix:
```suggestion
    reference: "bipolar", or None, default=None
```
I think you can remove this shortcut

neuralset-repo/neuralset/extractors/test_neuro.py
```diff
@@ -1458,6 +1459,155 @@ def test_bipolar_ref_validation() -> None:
...
+
+    extractor = ns.extractors.IeegExtractor(picks=("seeg",), reference="car")
+    result = extractor._preprocess_raw(
+        raw.copy(), tp.cast(etypes.MneRaw, SimpleNamespace(frequency=sfreq))
+    )
+    np.testing.assert_allclose(result.data.sum(axis=0), 0.0, atol=1e-5)
+    assert not np.allclose(result.data, 0.0)
+
+
+def test_ieeg_cannot_combine_car_and_bipolar_ref() -> None:
```
julien-gadonneix: Then you can also remove this test
Also one of you other test already check the mutual exclusiveness

neuralset-repo/neuralset/extractors/test_neuro.py
```diff
@@ -1458,6 +1459,155 @@ def test_bipolar_ref_validation() -> None:
...
+
+    extractor = ns.extractors.EegExtractor(picks=tuple(ch_names), car_ref=True)
+    result = extractor._preprocess_raw(
+        raw.copy(), tp.cast(etypes.MneRaw, SimpleNamespace(frequency=sfreq))
+    )
+    np.testing.assert_allclose(result.data.sum(axis=0), 0.0, atol=1e-5)
+    assert not np.allclose(result.data, 0.0)
+
+
+def test_car_ieeg_field_matches_shortcut() -> None:
```
julien-gadonneix: then you can also remove this test

neuralset-repo/neuralset/extractors/test_neuro.py
```diff
@@ -1458,6 +1459,155 @@ def test_bipolar_ref_validation() -> None:
...
+    """Regression guard: per-type CAR loop must skip channel types absent from raw."""
+    ch_names = ["A1", "A2", "A3"]
+    sfreq = 100.0
+    rng = np.random.RandomState(0)
+    data = rng.randn(len(ch_names), int(sfreq * 2)) + 5.0
+    info = mne.create_info(ch_names, sfreq=sfreq, ch_types="seeg")
+    raw = mne.io.RawArray(data, info)
+
+    # picks includes both seeg and ecog; raw has only seeg → ecog iteration must be skipped.
+    extractor = ns.extractors.IeegExtractor(picks=("seeg", "ecog"), reference="car")
```
julien-gadonneix:
```suggestion
    extractor = ns.extractors.IeegExtractor(picks=("seeg", "ecog"), car_ref=True)
```
then adapt this test

neuralset-repo/neuralset/extractors/test_neuro.py
```diff
@@ -1458,6 +1459,155 @@ def test_bipolar_ref_validation() -> None:
...
+    result = extractor._preprocess_raw(
+        raw.copy(), tp.cast(etypes.MneRaw, SimpleNamespace(frequency=sfreq))
+    )
+    np.testing.assert_allclose(result.data.sum(axis=0), 0.0, atol=1e-5)
+
+
+def test_car_eeg() -> None:
```
julien-gadonneix: this test can then also be removed

bentang18:
@julien-gadonneix  Done in `8c91970`:

- Removed the `IeegExtractor.reference="car"` shortcut (docstring, `Literal`, and `model_post_init` normalization).
- Deleted `test_ieeg_cannot_combine_car_and_bipolar_ref`, `test_car_eeg`, and `test_car_ieeg_field_matches_shortcut`.
- Switched the remaining `IeegExtractor` CAR tests to `car_ref=True`.

Also narrowed `Literal["bipolar", "car"]` to `Literal["bipolar"]` to match.

julien-gadonneix (approved):
Thank you !

PR #69, opened 2026-05-15
Switch Excel engine from openpyxl to calamine (~6× faster) in Nieuwland and Chen studies

juan-cobos:
This PR replaces the default openpyxl engine with calamine for `pd.read_excel` calls in `Nieuwland2018Large._preproc_stimuli` and `Chen2023Large._stimulus_info`, and merges Nieuwland's two separate sheet reads into one call (one file open instead of two). Also fixes a pre-existing typo in `Chen2023Large.requirements`: `"openxl"` → `"python-calamine"`.


kingjr:
thanks @juan-cobos

PR #70, opened 2026-05-15
Seed before neuralbench model construction

ducquoc97:
## Summary
- seed at the start of \ so model construction does not depend on prior RNG consumption
- keep the existing post-build reseed so training RNG stays decoupled from model-build RNG usage
- add a regression test and a standalone reproduction script for the wrapper/adapter/lazy-init path

## Why
In this codepath, "encoder only" does not mean model construction is RNG-free. RNG is still consumed by lazy-layer materialization, random channel-adapter initialization, and downstream wrapper probe construction. Seeding only after \ lets those parameters vary with earlier RNG usage such as data preparation.

## Verification
- \.                                                                        [100%]
1 passed in 8.84s
- \Late seeding digests:
  pre_draws=0  -> 52bfaf80b8aa8fc7
  pre_draws=64 -> e83d6513766c0cc0
Early seeding digests:
  pre_draws=0  -> 6898bd8e07026441
  pre_draws=64 -> 6898bd8e07026441
Reproduction succeeded: only early seeding is invariant to prior RNG usage.


hubertjb:
Thanks @ducquoc97, good catch! 

Could you please remove the `scripts` folder + content from the PR ?

Also, the new unit test fails type checking. A fix would be to add the `data` and `trainer_config` parameters, and call `typing.cast()` on all inputs initialized with `object()`. Happy to take care of it too, please let me know.

ducquoc97:
Thanks for your review @hubertjb , I've updated the code. Do you think we also need to seed when do the data setup? which mean seed once at the top of run(), before self.data.prepare(), with pl.seed_everything(self.seed, workers=True)?

hubertjb:
Thanks for the changes @ducquoc97. There were indeed some potential unseeded random calls (e.g. sampler and dataloader) which your last commit will catch too.

I'll merge as is, but fyi I'm working on a deeper refactor of rng seeding.

Thanks again for your contribution!

PR #79, opened 2026-05-21
Pack benchmark experiments into fewer scheduler jobs

bruAristimunha:
## Summary

`neuralbench` currently submits one Slurm job per experiment, which hits job-array limits on large grids and wastes scheduler overhead on short experiments that could share a node (see facebookresearch/exca#179). This PR adds opt-in batching: pending experiments share one scheduler job, with optional in-process parallelism inside the job. Default behavior is unchanged.

## What changed

- New CLI flags `--experiments-per-job N` (or `all`) and `--local-workers-per-job M`. Defaults `1`/`1` reproduce the previous one-job-per-experiment path.
- New `neuralbench.packing` module. The packed job invokes each inner experiment's cached `.run()`, so re-running the same grid *without* the flag still hits each individual exca cache.
- No new dependencies (stdlib `ProcessPoolExecutor` with `spawn`).

## Local scaling

Local benchmark, 8 synthetic experiments on a 14-core macOS machine, `ProcessPoolExecutor(mp_context="spawn")`:

| `--local-workers-per-job` | 1 | 2 | 4 | 8 |
|---|---|---|---|---|
| 10 s/exp (realistic) | 80.1 s | 45.6 s | 25.7 s | **17.4 s (4.6×)** |
| 2 s/exp (overhead-dominated) | 16.1 s | 13.8 s | 10.0 s | 9.1 s (1.8×) |

Spawn boot (~5 s/worker) amortizes for ≥10 s-per-experiment workloads.

## Known limitations

- Scheduler resource budget (CPU/GPU/mem/time/partition) is taken from the first pending experiment — group by resource hints before packing.
- No automatic `CUDA_VISIBLE_DEVICES` slicing with `--local-workers-per-job > 1`; keep it at `1` for GPU-bound experiments.

## Test plan

- [ ] `pytest neuralbench-repo/neuralbench/test_packing.py`
- [ ] Internal cluster: run with `--experiments-per-job 8`, then re-run without the flag — second run should submit zero new jobs.


bruAristimunha:
The benchmark results that motivate this PR are here:
pack_scaling

bruAristimunha:
I think the reviewers here are @jrapin, @kingjr, and @hubertjb. 

The core modification is described at https://github.com/facebookresearch/exca/issues/179.

neuralbench-repo/neuralbench/cli.py
```diff
@@ -42,6 +42,36 @@
 logger = logging.getLogger(__name__)
 
 
+def _parse_experiments_per_job(value: str) -> int | tp.Literal["all"]:
```
kingjr: shouldn't this be handled by pydantic validators?

neuralbench-repo/neuralbench/packing.py
```diff
@@ -0,0 +1,144 @@
...
+        if self.n_jobs == 1 or len(self.experiments) <= 1:
+            return [exp.run() for exp in self.experiments]
+
+        max_workers = min(self.n_jobs, len(self.experiments))
+        ctx = multiprocessing.get_context("spawn")  # CUDA-safe
+        # Manual lifecycle + ``as_completed`` so the first child failure
+        # surfaces immediately and ``shutdown(cancel_futures=True)`` aborts
+        # whatever is still pending — instead of waiting for the slowest
+        # in-flight task to drain before reporting.
+        pool = ProcessPoolExecutor(max_workers=max_workers, mp_context=ctx)
```
kingjr: shouldn't this be addressed with exca job array?

bruAristimunha: Fair question @kingjr! 

It is complementary, not an alternative. 

The motivation isn't parallelism only; it's about the scheduler-job count and getting as much of the GPU as possible. `job_array` still creates `N` independent Slurm tasks, each with its own allocation, queue slot, and fairshare/quota charge. 

On clusters with a concurrent-job cap (e.g., 200, like jeanzay) or a per-user GPU-hour budget eaten by submission overhead (like Access cluster), a 5000-cell grid doesn't fit through job_array alone. 

Packing 8-into-1 brings it down to ~625 tasks and frees the quota for other work.

neuralbench-repo/neuralbench/cli.py
bruAristimunha: Yes, good point. I am addressing!

bruAristimunha: hey @kingjr, done :)

bruAristimunha:
Hey @kingjr, should I transfer this to Exca? I didn't get any answer from @jrapin when I asked a few months ago.

PR #85, opened 2026-05-22
Fix digest type for SHA-256 in download.py

julien-gadonneix:
## What does this PR do?

Small fix when the digest type is incorrect in the files downloaded from Dryad for the conventions of `hashlib`.


kingjr:
LGTM, fix CI and I'll merge, thanks @julien-gadonneix !

PR #88, opened 2026-05-22
Fix HuggingFaceImage (and friends) for device="accelerate"

lucyzmf:
## What does this PR do?

`HuggingFaceImage(device="accelerate")` used to crash on the first
`tensor.to("accelerate")` call because `"accelerate"` is a HuggingFace
dispatch mode, not a real torch device. The same trap existed in
`HuggingFaceVideo`, and `_HFVideoModel` never propagated
`device_map="auto"` / fp16, making `accelerate` a silent no-op for
non-LLaVA video models.

This PR fixes the crash by **resolving the `"accelerate"` sentinel
exactly once, at construction time**, into normal torch state —
mirroring how `device="auto"` is already resolved in
`HuggingFaceMixin.model_post_init`. After init:

- `self.device` is always a real torch device (`"cuda"` or `"cpu"`).
  The `"accelerate"` string never leaks past the constructor.
- `self._hf_kwargs` (a single `pydantic.PrivateAttr`) carries the
  accelerate-specific `from_pretrained` kwargs —
  `{"device_map": "auto", "torch_dtype": torch.float16}` when
  dispatching, `{}` otherwise. Its truthiness is the single source of
  truth for "is accelerate dispatching this model?".

Image / text / video are rewired to consume only `self.device` and
`self._hf_kwargs`. There are no `device == "accelerate"` checks left in
any consumer.

## Behaviour changes (in the "code now matches the documented contract" direction)

- `HuggingFaceImage` with `device="accelerate"` no longer crashes and
  now actually dispatches the model across GPUs in fp16, matching
  `HuggingFaceText`'s pre-existing behaviour.
- `HuggingFaceVideo` honours `device="accelerate"` for all video
  models (previously only LLaVA-Video / LLaVA-34B used fp16 / dispatch,
  via hardcoded branches inside `_HFVideoModel`).
- `HuggingFaceText` external behaviour is unchanged; its `_load_model`
  and `_get_data` are refactored to use `self._hf_kwargs` /
  `self.device`.
- `extractor.device` reads as `"cuda"` (not `"accelerate"`) after
  constructing with `device="accelerate"`. This matches how
  `device="auto"` is already resolved at init and has no caching impact
  (see below).

## Latent bugs fixed along the way

- `HuggingFaceText._load_model` previously did
  `if accelerate: kwargs = {...}`, silently overwriting any caller
  kwargs. Now merged: `kwargs = {**self._hf_kwargs, **kwargs}`.
- `_HFVideoModel.__init__` previously did
  `extra = {"torch_dtype": ...}` and `extra = {"_attn_implementation": ...}`,
  each clobbering earlier branches. Now uses `extra[...] = ...` /
  `extra.update(...)` and seeds `extra` from the caller's `model_kwargs`.

## Cache UID / schema impact

None. `_hf_kwargs` is a `pydantic.PrivateAttr` — not a model field, so
it does not enter `model_dump()` or the cache UID. `device` is already
in `_exclude_from_cls_uid` and `_exclude_from_cache_uid`, so mutating it
from `"accelerate"` to `"cuda"` at init is invisible to caching. The
pinned `name=HuggingFaceImage-c5c2ebfa` assertion in `test_image.py`
still passes locally.

## Tests

- `test_image_accelerate_device_resolution` — CPU-safe regression
  asserting that `device="accelerate"` is resolved at init into
  `self.device == "cuda"` and the expected `_hf_kwargs` dict; also
  exercises `torch.empty(0).to(extractor.device)` when CUDA is
  available (the original crash path).
- `test_image_accelerate_end_to_end` — full extractor run on a >=2-GPU
  node, skipped otherwise.

## Known follow-ups (out of scope for this PR)

- `audio.py` has the same defect (5 `_model.to(self.device)` sites
  across `HuggingFaceAudio` + 4 subclasses, each with its own
  `_get_sound_model`, plus 1 `features.to(self.device)` in
  `_process_wav`). The same single-attribute pattern applies 1:1
  (`if not self._hf_kwargs: model.to(self.device)` + use `self.device`
  directly for tensor moves). 
- fp16 vs fp32 share the same cache key. `device` is excluded from
  both UIDs, so a `cuda` run (fp32) and an `accelerate` run (fp16) of
  the same extractor config now write to and read from the same cache
  file. Pre-existing in `HuggingFaceText`; this PR makes it observable
  in image and video too. A separate issue should decide whether to
  capture `dtype` in the cache UID or split `device` into "physical
  device" + "dtype policy".
- `device="accelerate"` on CPU-only systems still resolves to
  `self.device = "cuda"` and would fail at the first `tensor.to("cuda")`.
  Realistically not a use case; could be promoted to an explicit
  `model_post_init` validation error in a follow-up.


neuralset-repo/neuralset/extractors/text.py
```diff
@@ -492,8 +489,8 @@ def _get_data(
                     yield out
                 # erase variables / free memory
                 del hidden_states, hidden_state, word_state, states, outputs, inputs
-                if self.device == "accelerate" and torch.cuda.is_available():
-                    # in case of multi-GPU models, explicitly empty cache "just in case"
+                if self._hf_kwargs:
```
jrapin: the condition and the comment seem unrelated

neuralset-repo/neuralset/extractors/text.py
```diff
@@ -374,7 +373,8 @@ def _load_model(self, **kwargs: tp.Any) -> nn.Module:
             with torch.no_grad():
                 for p in model.parameters():
                     part_reversal(p)
-        if self.device != "accelerate":
+        if not self._hf_kwargs:
+            # accelerate dispatches the model internally; otherwise place it ourselves
```
jrapin: comment seems unrelated + huge border effect to having any hf_kwargs than could be anything but accelerate

neuralset-repo/neuralset/extractors/test_image.py
```diff
@@ -155,6 +155,38 @@ def test_image_token_aggregation(
     assert out.ndim == 2 if token_aggregation is None else 1
```
jrapin, line 157: read internal testing.md and coding-style.md and apply

neuralset-repo/neuralset/extractors/image.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         model_name: str,
         output_hidden_states: bool = False,
         pretrained: bool = True,
+        model_kwargs: dict[str, tp.Any] | None = None,
```
kingjr, line 64: I'm not a fan of this non-typed api, any more pydantic/neuralset like alternative?

sdascoli:
Discussed offline: remove the statefull fields _uses_accelerate (replace with a configurable field "use_accelerate") and _hf_kwargs (use a property instead)

neuralset-repo/neuralset/extractors/text.py
```diff
@@ -463,8 +460,7 @@ def _get_data(
                         return_tensors="pt",
                         padding=True,
                         truncation=True,  # beware to have set truncation_side="left" in init
-                        max_length=self._get_max_length(),  # guard tokenizers reporting no real limit (e.g. OPT)
-                    ).to(device)
+                    ).to(self.device)
```
sdascoli, line 464: Why remove max_length?

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -479,14 +481,25 @@ class HuggingFaceMixin(base.BaseModel):
...
+    device: tp.Literal["auto", "cpu", "cuda"] = "auto"
+    use_accelerate: bool = False
     layers: float | list[float] | tp.Literal["all"] = 2 / 3
     cache_n_layers: int | None = None
     layer_aggregation: tp.Literal["mean", "sum", "group_mean"] | None = "mean"
     token_aggregation: tp.Literal["first", "last", "mean", "sum", "max"] | None = "mean"
     _REPOS: tp.ClassVar[list[str]] = []
     _skip_repo_check: bool = False  # for simpler hacking (eg: custom dinov2 checkpoints)
 
+    @property
```
sdascoli, line 493: Rename model_build_kwargs

PR #98, opened 2026-05-26
Support full local neuralbench runs via --local

ducquoc97:
## What does this PR do?

Adds a `--local` mode to `neuralbench` so canonical runs can execute on the current machine without switching to the reduced debug config.

This is useful when the default non-debug path would otherwise rely on cluster execution. With `--local`, the benchmark and extractor/target infra are forced to run locally, while epochs, batch limits, and the rest of the task config stay unchanged.

Example:

```bash
neuralbench eeg audiovisual_stimulus --local
```

This branch also keeps W&B disabled when `WANDB_HOST` is blank, which matters for local runs that should not try to initialize W&B.

## Related Issues

N/A

## Checklist

- [x] Tests added or updated

### DIFF
--- a/neuralbench-repo/neuralbench/cli.py
+++ b/neuralbench-repo/neuralbench/cli.py
@@ -86,7 +86,7 @@ def run_benchmark(
     debug : bool
         Run locally with a reduced config (2 epochs, 5 batches).
     local : bool
-        Run locally without applying the reduced debug-mode config.
+        Run locally using the full config; unlike --debug, do not reduce epochs or batches.
     force : bool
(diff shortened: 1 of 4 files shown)


hubertjb:
Thanks for the PR @ducquoc97. I think this is a good feature to have, however I’d prefer not adding more flags to the CLI. I’m working on a short PR to make sure these can be controlled from the main config.json instead: #118 

Please let me know if this doesn’t cover your use case.

ducquoc97:
Thanks @hubertjb, sounds good to me

PR #99, opened 2026-05-26
Pass dataset channel names to custom model builds

ducquoc97:
## Summary
- pass dataset channel names to custom model configs that opt in with `ch_names_required`
- add regression coverage for the custom-model factory path

## Why
Custom models built through the non-braindecode factory path currently receive `n_in_channels` and `n_outputs`, but they are not able to access the dataset's actual channel names at build time.

That makes it hard for models that need channel identity, not just channel count, for tasks like validating expected inputs, mapping channels into a fixed layout, or initializing channel-specific logic.

This change keeps the behavior narrow and explicit: only models that declare `ch_names_required` receive `ch_names`, while existing models keep the current behavior.

## Testing
- `uv run --extra dev pytest neuralbench/test_model_factory.py`


hubertjb:
Thanks @ducquoc97, I’ll test a few things first and get back to you on this.

PR #103, opened 2026-05-26
Add 'voronoi' downsampling method to SurfaceProjector

sdascoli:
## Why this PR?
When we specify a mesh different from the input mesh, SurfaceProjector takes the first N vertices of each hemisphere. There are two important issues with this method:
- SNR loss (the most important): subsampling in this way means we loose a lot of SNR since we discard a lot of data. A better approach is to smooth over nearest neighbors
- Silent bug: doesn't work when going below fsaverage5 because of the vertex shuffling bug ([nilearn/nilearn#3415](https://github.com/nilearn/nilearn/issues/3415))

## Summary

Add a single new field on ``SurfaceProjector``:

```python
downsampling_method: Literal["prefix", "voronoi"] = "prefix"
```

that selects how 2-D surface data is downsampled to a coarser fsaverage mesh.

* **`"prefix"` (default — unchanged behaviour)**: keep the first `V_target` vertices of each hemisphere. fsaverage5/6/7 vertices are hierarchically nested by icosahedral subdivision, so the prefix slice returns the exact target-mesh vertices — but it discards the higher-resolution vertices instead of averaging them in, lowering SNR per retained vertex. A `UserWarning` is now emitted whenever this path actually shortens the input, recommending `"voronoi"`.
  Targeting `fsaverage3` or `fsaverage4` with this method now raises a `ValueError`: nilearn ships those two meshes with a vertex ordering that does *not* match fsaverage5+ ([nilearn/nilearn#3415](https://github.com/nilearn/nilearn/issues/3415)), so the prefix slice would return wrong values, not just lossy ones.
* **`"voronoi"` (new)**: average source vertices within each target Voronoi cell on the inflated mesh. Implementation: build a `cKDTree` over the *target* inflated mesh, query each *source* vertex with `k=1` to find its owner cell, then `np.add.at`-scatter source values into per-cell sums and divide by counts. Empty cells (which occasionally occur because nilearn's inflated meshes are not perfectly nested across fsaverage levels) fall back to the value of their nearest source vertex. Each Voronoi cell holds ~`4ⁿ` source vertices when downsampling `n` icosahedral levels.

The 4-D `vol_to_surf` path and the upsampling `NotImplementedError` are unchanged from `main`.

## Before / after

```python
# Before — identical to current main (still works, now with a warning)
SurfaceProjector(mesh="fsaverage5").apply(rec_fsaverage6)
# UserWarning: ...keeps the first N vertices...lowering SNR. Pass
# downsampling_method='voronoi' for a proper Voronoi-based resampling.

# New — Voronoi-based mean per target cell, more representative
SurfaceProjector(
    mesh="fsaverage5", downsampling_method="voronoi"
).apply(rec_fsaverage6)

# Targeting the broken-ordering meshes with prefix is now an explicit error
SurfaceProjector(mesh="fsaverage3").apply(rec_fsaverage7)
# ValueError: ...nilearn ships fsaverage3/4 with a different vertex
# ordering than fsaverage5+...Use downsampling_method='voronoi' instead.
```

## Test plan

`test_fmri_mesh_from_2d` is extended to cover:

- [x] `"prefix"` fsaverage6 → fsaverage5: warning emitted; output is exactly the first 10242 vertices per hemisphere of the input.
- [x] `"voronoi"` fsaverage6 → fsaverage5 via `FmriExtractor`: shape `(20484, 5)`, all finite.
- [x] `"voronoi"` constant input round-trips exactly (validates the empty-cell fallback).
- [x] Same source and target mesh: identity, no warning.

`pytest -k "fmri or surface"` → 12 passed, 1 skipped (pre-existing skip).

Co-authored-by: Cursor Agent <199161495+cursoragent@users.noreply.github.com>


kingjr:
I think prefix is actually a bug, no one using fs3/4/5 wants to compute a a small parcel without averaging

sdascoli:
> I think prefix is actually a bug, no one using fs3/4/5 wants to compute a a small parcel without averaging

Yeah I kind of agree, but @yohann-benchetrit subsequently convinced me that it's still better to have prefix by default, perhaps TBD during next tech meeting

sdascoli:
Just ran encode/decode tasks on NSD Betas

| Mode | Downsampling method | R |
|---|---|---:|
| encode | voronoi | 0.12320159737768187 |
| encode | prefix | 0.0986403038164134 |
| decode | voronoi | 0.3197358648625289 |
| decode | prefix | 0.2982897741926759 |

neuralbench-repo/neuralbench/defaults/config.yaml
```diff
@@ -13,7 +13,7 @@ wandb_config:
   project: !!python/name:neuralbench.config_manager.PROJECT_NAME
   name: null
 infra:
-  cluster: auto
+  cluster: !!python/name:neuralbench.config_manager.CLUSTER
```
kingjr, line 16: are all these changes expected?

kingjr:
closing in favour of Chain approach (incoming)

PR #106, opened 2026-05-26
Revert and re-attribute neuralbench studies

hubertjb:
The purpose of this PR is to fold contributor attribution back into the public history via Co-authored-by trailers - as a result the diff is intentionally empty.


teonbrooks:
LGTM. thanks @hubertjb!

PR #108, opened 2026-05-28
Auto-inject sfreq + chs_info so all braindecode models build

bruAristimunha:
## Summary

`build_braindecode_model` previously forwarded only `n_chans`, `n_times`, `n_outputs` to the braindecode constructor (plus `chs_info` for FMs that opted in via `chs_info_required`). Most classifier-shaped braindecode classes also need `sfreq` to construct — without it, ~13 of them (ContraWR, DGCNN, EEGMiner, EEGSimpleConv, EEGSym, FBMSNet, IFNet, SCCNet, ...) raised at `__init__`.

This change:
- Derives `sfreq` (and `ch_names`) from the neuro extractor once and forwards them alongside the existing args. `input_window_seconds` is intentionally not passed — braindecode auto-derives it from `n_times / sfreq`.
- Filters the resulting kwargs against the target class's `__init__` signature so models that don't accept some kwarg (e.g. TCN rejects `n_times`) still build. Respects `**kwargs` (a model whose `__init__` accepts var-keyword takes all kwargs verbatim).
- Lets `BaseBrainDecodeModel.build()` raise on overlap between runtime-derived and `config.kwargs` so data-derived values can't be silently overridden.

A new test in `neuralbench/test_model_factory.py` drives the catalogue from `braindecode.models.util.models_dict` and asserts 40 classifier-shaped models build and forward against a `(2, 22, 1000)` input. The `ALREADY_IMPLEMENTED` exclusion set is derived from the YAML configs in `neuralbench/models/` so it stays in sync as new FM YAMLs land; `NON_EEG_CLASSIFIERS` is a 6-item hand-maintained list for sleep stagers / EMG / MEG / self-supervised nets.

A second test pins the overlap-raise contract: if a YAML pre-sets one of the auto-injected keys (e.g. `kwargs: {sfreq: 250.0}`), `build_braindecode_model` must raise with `"kwargs overlap with config kwargs for keys"` rather than silently override.

## Test plan

- [x] `pytest neuralbench/test_model_factory.py` — 41 passed in ~1.5 s on `braindecode==1.6.0dev0`
- [x] Existing FM YAMLs (BENDR / BIOT / CBraMod / Labram / LUNA / REVE) still build via the `from_pretrained_name` branch — `NtBendr.build()`, `NtLabram.build()` etc. absorb any conflict-prone kwargs before delegating to `super().build()`
- [x] End-to-end runs on Expanse for the previously-failing classics (SCCNet, DGCNN, ShallowFBCSPNet, Deep4Net, EEGConformer, ...) train through the standard `neuralbench eeg <task> -m <model>` pipeline


bkowshik:
Tried this on a clean Colab as an independent check — a fresh `pip install` of the branch against the released `braindecode==1.5.2` (rather than the `1.6.0dev0` in the test plan). The three representative classes flip cleanly:

| model | `main` | this branch |
|---|---|---|
| `SCCNet` | `sfreq could not be inferred` | ✅ builds + forwards |
| `ContraWR` | `input_window_seconds could not be inferred` | ✅ |
| `DGCNN` | `chs_info not specified` | ✅ |

Nice that the three failure modes map exactly onto the three injected params (`sfreq`, `sfreq` → `input_window_seconds`, `chs_info`). Confirms it works on the released braindecode too, not just `1.6.0dev0`. 👍

<details>
<summary>Minimal repro (Colab)</summary>

```bash
B="git+https://github.com/bruAristimunha/neuroai.git@feat/auto-eeg-params"
pip install "$B#subdirectory=neuralset-repo" "$B#subdirectory=neuraltrain-repo" "$B#subdirectory=neuralbench-repo" x_transformers
# swap to facebookresearch/neuroai.git@main to see the failures
```

```python
from types import SimpleNamespace
import torch
from neuraltrain.models import base as bd_base
from neuralbench.model_factory import build_braindecode_model

CH = ["Fp1","Fp2","F3","F4","C3","C4","P3","P4","O1","O2","F7","F8",
      "T7","T8","P7","P8","Fz","Cz","Pz","FC1","FC2","CP1"]
loader = SimpleNamespace(dataset=SimpleNamespace(
    extractors={"neuro": SimpleNamespace(frequency=120.0, _channels=dict.fromkeys(CH))}))

for name in ["SCCNet", "ContraWR", "DGCNN"]:
    try:
        m = build_braindecode_model(getattr(bd_base, name)(), None, loader, 22, 1000, 4)
        m.eval(); out = m(torch.randn(2, 22, 1000))
        print(f"{name:10} -> OK {tuple(out.shape)}")
    except Exception as e:
        print(f"{name:10} -> FAIL {str(e).splitlines()[0]}")
```

</details>

neuralbench-repo/neuralbench/model_factory.py
```diff
@@ -82,6 +86,9 @@ def build_braindecode_model(
     else:
         build_kwargs.update(n_chans=n_in_channels, n_times=n_times)
 
+    params = inspect.signature(type(brain_model_config)._MODEL_CLASS.__init__).parameters
```
kingjr: anything cleaner we could do?

neuralbench-repo/neuralbench/model_factory.py
```diff
@@ -60,13 +60,17 @@ def build_braindecode_model(
...
     if _needs_ch_names:
-        neuro_extractor = train_loader.dataset.extractors["neuro"]  # type: ignore[attr-defined]
-        ch_names = list(neuro_extractor._channels.keys())
         assert len(ch_names) == n_in_channels, (
             f"Expected {n_in_channels} channels, but got {len(ch_names)} channel names."
         )
+    if len(ch_names) == n_in_channels:
         build_kwargs["chs_info"] = [{"ch_name": name} for name in ch_names]
+    frequency = getattr(neuro_extractor, "frequency", None)
+    if isinstance(frequency, (int, float)):
```
kingjr, line 75: maybe use ns.Frequency

neuralbench-repo/neuralbench/model_factory.py
bruAristimunha: good provocation @kingjr... I think we can maybe use the config approach from braindecode, let me check....

neuralbench-repo/neuralbench/model_factory.py
bruAristimunha: I will check if the ns.Frequency is updated once we resample or apply other operations

PR #112, opened 2026-05-29
Save recording-level test predictions to disk

bruAristimunha:
## What

Adds optional on-disk saving of recording-level test predictions to the
`RecordingLevelEval` callback. When a `save_dir` is provided, after computing
per-recording averaged predictions the callback writes
`<save_dir>/predictions/test_recording_level.pt` containing, per timeline:
`y_true`, per-class probabilities (`y_pred_prob{i}`), the averaged per-class
scores (`y_pred{i}`), and the arg-max label (`y_pred`).

## Why

Downstream analysis (per-dataset breakdowns, confusion matrices, layer/model
comparison tables, calibration) needs the raw recording-level predictions, not
just the aggregate metric logged to the trainer. Persisting them once at eval
time avoids re-running inference and makes results reproducible from cache.

## Changes

- `neuralbench/callbacks.py` — `RecordingLevelEval.__init__(save_dir=None)`;
  new `_save_outputs()` helper; also exposes `y_pred` (arg-max) and
  `y_pred_prob{i}` columns alongside the existing `y_pred{i}` scores.
- `neuralbench/main.py` — wire `save_dir` through when constructing the callback.
- `neuralbench/test_callbacks.py` — coverage for the save path (file written,
  schema, no-op when `save_dir is None`).

## Notes

- Default `save_dir=None` → fully backward compatible (no file written, no
  behavior change for existing runs).
- Output format is a plain dict-of-lists via `torch.save`, loadable without any
  project imports.


bruAristimunha:
Close in favor of @hubertjb approach

PR #113, opened 2026-05-29
Add audio word extraction transform

sdascoli:
## Summary
- Add `ExtractWordsFromAudio`, a WhisperX-backed events transform that turns `Audio` rows into aligned `Word` rows.
- Cache transcripts as sidecar `.tsv` files, reuse existing caches unless `overwrite=True`, and preserve audio-row metadata on generated words.
- Cover fresh transcription and cached-transcript behavior with focused unit tests.

## Test plan
- `.venv/bin/python -m pytest neuralset-repo/neuralset/events/transforms/test_transforms.py -k 'extract_words_from_audio'`
- `.venv/bin/python -m ruff check neuralset-repo/neuralset/events/transforms/audio.py neuralset-repo/neuralset/events/transforms/test_transforms.py`
- Full `test_transforms.py` run: 86 passed, 1 environment failure due missing `deepmultilingualpunctuation` for `test_ensure_texts_fullstop`.

### DIFF
--- a/neuralset-repo/neuralset/events/transforms/audio.py
+++ b/neuralset-repo/neuralset/events/transforms/audio.py
@@ -181,17 +181,19 @@ class ExtractWordsFromAudio(EventsTransform):
             "offset",
             "stop",
         }
-        for audio_event in audio_events.itertuples(index=False):
-            transcript = transcripts[str(Path(audio_event.filepath))].copy(deep=True)
+        for _, audio_event in audio_events.iterrows():
+            audio_filepath = Path(tp.cast(str, audio_event["filepath"]))
+            transcript = transcripts[str(audio_filepath)].copy(deep=True)
             if transcript.empty:
                 continue
-            for key, value in audio_event._asdict().items():
+            for key, value in audio_event.items():
                 if key not in ignored_fields:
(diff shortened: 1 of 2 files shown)


kingjr:
This should be in experimental/, we'll need proper benchmark before moving to a public api.

sdascoli:
This is what is used in Tribe, I think it’s quite mature (been using over a year), are you sure?
https://github.com/facebookresearch/tribev2/blob/main/tribev2/eventstransforms.py

kingjr:
For now this looks like research code, not public library
- This is using a subprocess (not pure python), 
- it assumes having uvx but package supports conda too
- whisperx often generate poor transcripts
- there is a try except pattern that should be avoided
- device is not parameterizable

sdascoli:
Yeah the subprocess and uv is to avoid having to install whisper which litterally breaks all environments… I think it’s necessary
I can fix the other things but happy to drop the PR if you think that’s better

kingjr:
Yes, I'm afraid this is the kind of environment issue makes it ill-suited for support here. There is also efforts from Andrea to build a systematic transcription pipeline, and would prefer having a single clear object that we verified on all of our onboarded speech studies.

PR #114, opened 2026-05-29
Support triggerless strided segmentation

sdascoli:
## Summary
- Allows `Segmenter(trigger_query=None, stride=...)` to build strided windows without trigger events.
- Creates triggerless windows over each full timeline, from that timeline's minimum start to maximum stop, while preserving the existing `start` offset and `stride_drop_incomplete` behavior.
- Keeps all events available in triggerless mode so timeline bounds are not narrowed by extractor event-type filtering.

## Example
```python
segmenter = ns.Segmenter(
    extractors={"meg": ns.extractors.MegExtractor(frequency="native")},
    trigger_query=None,
    duration=2.0,
    stride=0.5,
)
dset = segmenter.apply(events)
```

## Test plan
- `./.venv/bin/python -m pytest neuralset-repo/neuralset/test_segments.py::test_list_segments_stride_without_triggers neuralset-repo/neuralset/test_dataloader.py::test_segmenter_stride_without_trigger_query`
- `./.venv/bin/python -m pytest neuralset-repo/neuralset/test_segments.py`

### DIFF
--- a/neuralset-repo/neuralset/dataloader.py
+++ b/neuralset-repo/neuralset/dataloader.py
@@ -477,7 +477,8 @@ class Segmenter(base.BaseModel):
         Dataframe query selecting which events act as :term:`triggers <trigger>`
         — segments are time-locked to the matching events
         (see :data:`base.Query`).
-        At least one of ``trigger_query`` or ``stride`` must be provided.
+        If None, ``stride`` must be provided and strided segments are created
+        across each full timeline.
     stride: optional float
         Stride (in seconds) to use to define sliding window segments.
     stride_drop_incomplete: optional bool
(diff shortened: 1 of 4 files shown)


kingjr:
I don't understand what functionality this adds beyond not having to specify e.g. trigger=Fmri

sdascoli:
Yes sorry forgot to explain in the description. 
The reason I introduce this is that there are case where we don't have events spanning the whole timeline:
- Timelines where the neuro is split into chunks
- In-silico experiments for Tribe

In such cases, it's cumbersome to make strided windows across the timeline: the only solution is to hack by creating fake events which span each timeline fully, which doesn't seem like a clean approach

kingjr:
I'm afraid of the added complexity and downstream consequences e.g. on dset.subselect and thus train/test splitting / subject dispatching.

In those rare cases neuros are chunked, can we fix this by adding a Block event that spans the whole thing>

PR #115, opened 2026-05-29
feat(segments): add on_trigger_overlap guard to list_segments

kingjr:
## Summary

Adds an `on_trigger_overlap` parameter to `list_segments` that, **by default, raises** when two or more trigger events of the **same `type`** start at the **same time** on the **same `timeline`** (e.g. two `Fmri` events differing only by metadata such as `space`/`preproc`).

Such simultaneous, similarly-typed triggers silently produce duplicate/ambiguous segments and can corrupt train/test splits — the failure mode discussed in fairinternal/brainai#2531 (comment: https://github.com/fairinternal/brainai/pull/2531#issuecomment-4073936157):

> 1. `list_segments(on_trigger_overlap='raise')` to ensure a single trigger, avoid bug with stride

The guard inspects only the **trigger** events (`events.loc[triggers]`), so legitimately overlapping *context* events are unaffected. It runs *before* the strided/non-strided branch, so it covers both modes (catching the "avoid bug with stride" case where colliding triggers would each expand into the same set of windows).

## API

```python
list_segments(events, triggers, *, on_trigger_overlap="raise" | "warn" | "allow")
```

- `"raise"` (default): error, printing the colliding `(type, timeline, start)` rows.
- `"warn"`: log a warning and still build the (duplicate) segments.
- `"allow"`: disable the check (previous behavior).

## Note on the default

Defaulting to `"raise"` is a deliberate behavior change matching the "(by default)" intent in the linked comment. It surfaces pipelines that build segments from simultaneous same-type triggers (e.g. multi-space `Fmri`); those should narrow the `triggers` mask or pass `on_trigger_overlap="allow"`. Happy to flip to `"warn"` if we'd rather phase it in.

Open question (flagged in review): should the collision key drop `type` and use only `(timeline, start)`? For the common single-type trigger this is equivalent; keeping `type` avoids false positives on intentional multi-type triggering where the trigger event carries meaning (e.g. `aggregation="trigger"`).

## Test plan

- [x] New unit test `test_on_trigger_overlap` in `test_segments.py`: covers `raise` (default + explicit), `warn` (logs + builds), `allow`.
- [x] ruff / codespell clean; mypy clean on `segments.py`.
- [ ] CI (lint, mypy, pytest neuralset).

Made-with: Cursor

### DIFF
--- a/neuralset-repo/neuralset/segments.py
+++ b/neuralset-repo/neuralset/segments.py
@@ -475,10 +475,6 @@ def _check_trigger_overlap(
     """
     if mode == "allow":
         return
-    if mode not in ("raise", "warn"):
-        raise ValueError(
-            f"on_trigger_overlap must be 'raise', 'warn' or 'allow', got {mode!r}"
-        )
 
     keys = ["type", "timeline", "start"]
     if not all(k in trigger_df.columns for k in keys):
--- a/neuralset-repo/neuralset/segments.py
+++ b/neuralset-repo/neuralset/segments.py
@@ -332,7 +332,6 @@ def list_segments(
     duration: float | None = None,
     stride: float | None = None,
     stride_drop_incomplete: bool = True,
-    on_trigger_overlap: tp.Literal["raise", "warn", "allow"] = "raise",
 ) -> list[Segment]:
     """Create a list of segments based on events, sliding windows, or both.
 
@@ -353,9 +352,6 @@ def list_segments(
     stride_drop_incomplete : bool, optional
         If True and stride is not None, drop segments not fully contained within
         the valid time range. Default is True.
-    on_trigger_overlap : {"raise", "warn", "allow"}, optional
-        Behavior when several triggers of the same ``type`` start at the same
-        time on the same ``timeline`` (ambiguous segments). Default is "raise".
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


neuralset-repo/neuralset/segments.py
```diff
@@ -398,6 +402,23 @@ def list_segments(
     if not len(trigger_df):
         raise ValueError("Empty trigger events")
 
+    # Two triggers of the same type starting at the same time on the same
+    # timeline (e.g. two Fmri events differing only by metadata) define
+    # ambiguous/duplicate segments and can corrupt train/test splits. See
+    # https://github.com/fairinternal/brainai/pull/2531#issuecomment-4073936157
```
yohann-benchetrit: Reference public issue or remove

neuralset-repo/neuralset/segments.py
```diff
@@ -398,6 +402,23 @@ def list_segments(
...
+    if on_trigger_overlap != "allow":
```
yohann-benchetrit: Since `list_segments` is a plain function, `Literal` will not reject typos from scripts/configs; any value other than "raise"/"allow" behaves like "warn" when collisions exist.

yohann-benchetrit:
```suggestion
    valid_on_trigger_overlap = ("raise", "warn", "allow")
    if on_trigger_overlap not in valid_on_trigger_overlap:
        raise ValueError(
            f"on_trigger_overlap must be one of {valid_on_trigger_overlap}, "
            f"got {on_trigger_overlap!r}"
        )
    if on_trigger_overlap != "allow":
```

yohann-benchetrit (changes requested):

neuralset-repo/neuralset/segments.py
kingjr: good point. We can handle this with a query validator: https://github.com/facebookresearch/neuroai/blob/main/neuralset-repo/neuralset/base.py#L103C51-L103C65

yohann-benchetrit: Not really - If `on_trigger_overlap` remains only a `list_segments` argument, we can’t rely on the Pydantic validator pattern: `list_segments` is a plain function, so Literal/Pydantic validation will not run for direct calls. We could additionally expose it as a `Segmenter` attribute to validate config-based usage, but direct `list_segments` calls would still need an explicit runtime guard anyway.

kingjr: Is there a usecase where we need to allow a trigger_overlap?

kingjr: Do we have a usecase where one would need to have trigger_overlap? Else we don't expose the option and only raise.

yohann-benchetrit (approved):
LGTM, thanks @kingjr  !

PR #117, opened 2026-06-01
Avoid re-standardizing cached events

sdascoli:
## Summary
- Stop re-running `standardize_events` when loading `ValidatedParquet` caches; loads now return the cached parquet frame unchanged.
- Keep the derived `stop` column when dumping standardized events instead of dropping it and relying on load-time normalization to recreate it.


bkowshik:
Reviewed and verified this. LGTM.

The change does what it says: dump now persists `stop`, load returns the parquet frame unchanged (byte-identical round-trip), and the `numpy_nullable`→`pd.NA` guard the test documents still holds. Removing the load-time `standardize_events` is a nice win — it also drops a load path that could `raise` via `_propagate_bids`, so the cache becomes a transparent, can't-fail read.

PR #122, opened 2026-06-03
Fix duplicate Sentence rows in AddSentenceToWords with multiple Text events

julien-gadonneix:
## What does this PR do?

AddSentenceToWords called `_extract_sentences` inside the loop over `Text` contexts, so each pass re-derived sentences from the full `events` table and appended them again. Timelines with several `Text` rows could end up with duplicate `Sentence` events.
Sentence extraction now runs once after all contexts have been matched to words.

## Checklist

- [x] Tests added or updated


kingjr:
ouch thanks @julien-gadonneix !
cc @shouhamdi @BelCorentin @jrapin

PR #128, opened 2026-06-05
Add study folder name overrides

sdascoli:
## Issue Addressed

Studies that subclass another study can share data under the parent study folder. Before this change, `Study.model_post_init()` always looked for a subfolder named after `child.__class__.__name__`.
Example: Bel2026Petit has two subclasses Bel2026PetitRead and Bel2026PetitListen, but the data is shared in the same folder, Bel2026Petit.
When instantiating Bel2026PetitRead, Study looks for data in the Bel2026PetitRead instead of Bel2026Petit.
Our solution to this was to manually create a symlink Bel2026PetitRead->Bel2026Petit.

Issues with this symlink approach:
- Requires manual intervention for each new subclass and makes study folder messy
- Shared stimuli between the children will have different paths, which will force cache recomputation

## Summary
- Add a `Study.folder_name` class attribute. When not specified, fallback to class.__name__. 
- This folder name is used both for automatic folder discovery AND for study.download().
- This PR adds the folder_name for known parent classes such Bel2026Petit, TUH etc.

### DIFF
--- a/neuralfetch-repo/neuralfetch/studies/tuh_eeg.py
+++ b/neuralfetch-repo/neuralfetch/studies/tuh_eeg.py
@@ -678,8 +678,6 @@ class HaratiAbhishaike2015Tuev(Harati2015Tuev):
     reused in BIOT, LaBraM, CBraMod, etc.
     """
 
-    folder_name: tp.ClassVar[str | None] = "Harati2015Tuev"
-
     _info: tp.ClassVar[study.StudyInfo] = study.StudyInfo(
         num_timelines=518,
         num_subjects=370,
--- a/neuralfetch-repo/neuralfetch/studies/tuh_eeg.py
+++ b/neuralfetch-repo/neuralfetch/studies/tuh_eeg.py
@@ -45,7 +45,6 @@ import pandas as pd
 from neuralfetch import utils
 from neuralset.events import study
 from neuralset.events.etypes import CategoricalEvent
-from neuralset.events.study import _identify_study_subfolder
 
 logger = logging.getLogger(__name__)
 
@@ -110,6 +109,7 @@ def _fix_ch_name(name: str) -> str:
 
 
 class _BaseTuhEeg(study.Study):
+    folder_name: tp.ClassVar[str | None] = "tuh_eeg"
     aliases: tp.ClassVar[tuple[str, ...]] = ("TUH EEG Corpus", "TUH EEG")
     STUDY_CODE_MAP: tp.ClassVar[dict] = {
         "obeid2016": "tueg",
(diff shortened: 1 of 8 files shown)


kingjr:
How about making this fully automatic

```
class Study(Step):

    @classmethod
    def _folder_name(cls):
        base = max(
            (
                c for c in cls.mro()
                if issubclass(c, Study) and c not in (Study, cls)
            ),
            key=lambda c: len(c.mro()),
            default=cls,
        )

        return (
            base.__name__
            .lstrip("_")
            .removesuffix("Base")
        )
```

or make it depend on the name of the .py

neuralfetch-repo/neuralfetch/studies/tuh_eeg.py
kbegany, line 134: Is this still needed?

neuralfetch-repo/neuralfetch/studies/tuh_eeg.py
```diff
@@ -110,6 +109,7 @@ def _fix_ch_name(name: str) -> str:
 
 
 class _BaseTuhEeg(study.Study):
+    folder_name: tp.ClassVar[str | None] = "tuh_eeg"
```
kbegany, line 112: Does this work? Sub-studies need 
`NEURALSET_STUDY_FOLDER / "tuh_eeg" / {study_code}`, where `study_code` could be defined using either `STUDY_CODE_MAP` or `aliases`. There isn't a `download` subfolder, not sure if that could impact anything.

neuralset-repo/neuralset/events/study.py
```diff
@@ -308,6 +309,7 @@ class Study(base.Step):
     bibtex: tp.ClassVar[str] = ""
     licence: tp.ClassVar[str] = ""
     description: tp.ClassVar[str] = ""
+    folder_name: tp.ClassVar[str | None] = None
```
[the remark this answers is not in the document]
kbegany: Good idea

neuralset-repo/neuralset/events/study.py
```diff
@@ -428,14 +430,17 @@ def _to_timeline_string(self, timeline: dict[str, tp.Any]):
     def _exclude_from_cls_uid(cls) -> list[str]:
         return super()._exclude_from_cls_uid() + ["path"]
 
+    def _folder_name(self) -> str:
```
kbegany, line 433: What happens if user inputs a path, expecting to ignore `NEURALSET_STUDY_FOLDER` env variable? E.g., if study is in different location, not just has a different name than the classname

kbegany (changes requested):

sdascoli:
> How about making this fully automatic
> 
> ```
> class Study(Step):
> 
>     @classmethod
>     def _folder_name(cls):
>         base = max(
>             (
>                 c for c in cls.mro()
>                 if issubclass(c, Study) and c not in (Study, cls)
>             ),
>             key=lambda c: len(c.mro()),
>             default=cls,
>         )
> 
>         return (
>             base.__name__
>             .lstrip("_")
>             .removesuffix("Base")
>         )
> ```
> 
> or make it depend on the name of the .py

Yeah I was also considering this, in the end I found the remove suffix and lstrip a bit hacky… but happy either way

kbegany (approved):
LGTM see suggested points of revision (e.g., tuh_eeg)

kingjr (changes requested):
use code to detect folder name, rather than rely on a user defined property.

PR #135, opened 2026-06-08
Unify HuggingFace model loading and more

sdascoli:
## Why this PR?

- Model build, device and dtype handling is extremely messy and ad-hoc in our HuggingFace extractors, and different across modalities.  
- For some HF models, AutoModel/AutoProcessor do not work, you need to load a specific model class such as VivitModel. Currently this behaviour is hardcoded, meaning we only support a fraction of HF models
- We currently do not support giving arguments such as trust_remote_code to the HF model build, which forced people such as @shouhamdi to create custom subclasses for each edge case
- device="accelerate" was an alias for {"device_map": "auto" and "dtype=float16"}, which is quite arbitrary and caused bugs in HuggingFaceImage (see https://github.com/facebookresearch/neuroai/pull/88)

## Summary

Unifies HuggingFace model loading across text, audio, image, and video extractors by moving shared model construction options into a common `hf_config` and centralizing model instantiation in `HuggingFaceMixin`. This gives us full flexibility on model instantiation

## What Changed

- Added `HuggingFaceMixin.load_model()` and shared model/processor caches so text, audio, image, and video extractors use the same HuggingFace loading path.
- Restored explicit `device` and `dtype` fields on `HuggingFaceMixin`: `device="auto"` resolves to CUDA when available otherwise CPU, while `device="accelerate"` is the opt-in path that forwards `device_map="auto"` to HuggingFace.
- Added `HuggingFaceConfig` for HuggingFace class selection and escape-hatch kwargs: `model_cls_name`, `processor_cls_name`, `model_kwargs`, and `processor_kwargs`.
- Added public `HuggingFaceConfig.HF_CLASS_DEFAULTS` so known model-name patterns can resolve non-standard model/processor classes without modality-specific `if model_name` branches.
- Treat audio feature extractors and text tokenizers as shared HuggingFace processors, so modality-specific processor kwargs and caching follow the same path.
- Simplified image and video extractors by removing wrapper classes, merging `BaseImage` into `HuggingFaceImage`, and removing `_HFVideoModel`, `layer_type`, xclip-specific handling, and model-name-based frame defaults.
- Made native `HuggingFaceVideo.num_frames` required and warn when it differs from `model.config.num_frames` when available.
- Kept HuggingFace repo existence validation and added validation errors for misplaced `device_map`/`torch_dtype` entries in `hf_config.model_kwargs`; users should use `device`/`dtype` instead.

### DIFF
--- a/neuralset-repo/neuralset/events/transforms/test_transforms.py
+++ b/neuralset-repo/neuralset/events/transforms/test_transforms.py
@@ -1066,7 +1066,7 @@ def test_similarity_splitter(image_fpaths: list[Path]) -> None:
     )
 
     feat = ns.extractors.image.HuggingFaceImage(
-        hf_config={"device_map": "cpu"},  # type: ignore[arg-type]
+        device="cpu",
         layers=1.0,
     )
     splitter = _transf.SimilaritySplitter(extractor=feat)
--- a/neuralset-repo/neuralset/events/transforms/test_transforms.py
+++ b/neuralset-repo/neuralset/events/transforms/test_transforms.py
@@ -1066,7 +1066,8 @@ def test_similarity_splitter(image_fpaths: list[Path]) -> None:
     )
 
     feat = ns.extractors.image.HuggingFaceImage(
-        hf_config={"device_map": "cpu"}, layers=1.0
+        hf_config={"device_map": "cpu"},  # type: ignore[arg-type]
+        layers=1.0,
     )
     splitter = _transf.SimilaritySplitter(extractor=feat)
     events = splitter(mock_events)
--- a/neuralset-repo/neuralset/events/transforms/test_transforms.py
+++ b/neuralset-repo/neuralset/events/transforms/test_transforms.py
@@ -1065,10 +1065,7 @@ def test_similarity_splitter(image_fpaths: list[Path]) -> None:
         }
     )
 
-    feat = ns.extractors.image.HuggingFaceImage(
-        device="cpu",
-        layers=1.0,
-    )
+    feat = ns.extractors.image.HuggingFaceImage(device="cpu", layers=1.0)
     splitter = _transf.SimilaritySplitter(extractor=feat)
     events = splitter(mock_events)
 
--- a/neuralset-repo/neuralset/events/transforms/test_transforms.py
+++ b/neuralset-repo/neuralset/events/transforms/test_transforms.py
@@ -1065,7 +1065,9 @@ def test_similarity_splitter(image_fpaths: list[Path]) -> None:
         }
     )
 
-    feat = ns.extractors.image.HuggingFaceImage(hf_config={"device": "cpu"}, layers=1.0)
+    feat = ns.extractors.image.HuggingFaceImage(
+        hf_config={"device_map": "cpu"}, layers=1.0
+    )
     splitter = _transf.SimilaritySplitter(extractor=feat)
     events = splitter(mock_events)
 
--- a/neuralset-repo/neuralset/events/transforms/test_transforms.py
+++ b/neuralset-repo/neuralset/events/transforms/test_transforms.py
@@ -1065,7 +1065,7 @@ def test_similarity_splitter(image_fpaths: list[Path]) -> None:
         }
     )
 
-    feat = ns.extractors.image.HuggingFaceImage(device="cpu", layers=1.0)
+    feat = ns.extractors.image.HuggingFaceImage(hf_config={"device": "cpu"}, layers=1.0)
     splitter = _transf.SimilaritySplitter(extractor=feat)
     events = splitter(mock_events)
 
--- a/neuralset-repo/neuralset/extractors/audio.py
+++ b/neuralset-repo/neuralset/extractors/audio.py
@@ -338,12 +338,14 @@ class HuggingFaceAudio(BaseAudio, HuggingFaceMixin):
 
     model_name: str = "facebook/wav2vec2-large-xlsr-53"
     requirements: tp.ClassVar[tuple[str, ...]] = ("transformers>=4.29.2",)
+    hf_config: HuggingFaceConfig = HuggingFaceConfig(
+        processor_cls_name="AutoFeatureExtractor",
+    )
 
     normalized: bool = True
     layer_type: tp.Literal["transformer", "convolution"] = "transformer"
     # internal
     _model: nn.Module
-    _feature_extractor: nn.Module
 
     def model_post_init(self, log__: tp.Any) -> None:
         super().model_post_init(log__)
@@ -362,7 +364,9 @@ class HuggingFaceAudio(BaseAudio, HuggingFaceMixin):
 
     @property
     def _input_frequency(self) -> float:
-        return self.feature_extractor.sampling_rate  # type: ignore
+        processor = self.processor
+        feature_extractor = getattr(processor, "feature_extractor", processor)
+        return feature_extractor.sampling_rate  # type: ignore
 
     @classmethod
     def _exclude_from_cls_uid(cls) -> list[str]:
@@ -374,23 +378,14 @@ class HuggingFaceAudio(BaseAudio, HuggingFaceMixin):
         return base + HuggingFaceMixin._exclude_from_cache_uid(self)
 
     @property
-    def feature_extractor(self) -> nn.Module:
-        if not hasattr(self, "_feature_extractor"):
-            self._feature_extractor = self._get_feature_extractor(self.model_name)
-        return self._feature_extractor
+    def feature_extractor(self) -> tp.Any:
+        return self.processor
 
-    def _get_feature_extractor(self, model_name: str) -> torch.nn.Module:
-        from transformers import AutoFeatureExtractor
-
-        return AutoFeatureExtractor.from_pretrained(
-            model_name, **self.hf_config.config_build_kwargs
-        )
-
-    def _get_features(self, wav):
-        out = self._feature_extractor(
+    def _get_features(self, wav: torch.Tensor) -> tp.Any:
+        out = self.processor(
             wav,
             return_tensors="pt",
-            sampling_rate=self.feature_extractor.sampling_rate,
+            sampling_rate=self._input_frequency,
             do_normalize=self.normalized,
         )
         try:
@@ -466,6 +461,7 @@ class Wav2VecBert(HuggingFaceAudio):
     model_name: str = "facebook/w2v-bert-2.0"
     hf_config: HuggingFaceConfig = HuggingFaceConfig(
         model_cls_name="Wav2Vec2BertModel",
+        processor_cls_name="AutoFeatureExtractor",
     )
 
 
@@ -487,6 +483,7 @@ class SeamlessM4T(HuggingFaceAudio):
     model_name: str = "facebook/hf-seamless-m4t-medium"
     hf_config: HuggingFaceConfig = HuggingFaceConfig(
         model_cls_name="SeamlessM4TModel",
+        processor_cls_name="AutoFeatureExtractor",
     )
 
     def load_model(self) -> torch.nn.Module:
@@ -515,6 +512,7 @@ class Whisper(HuggingFaceAudio):
     hf_config: HuggingFaceConfig = HuggingFaceConfig(
         torch_dtype="float32",
         model_cls_name="WhisperModel",
+        processor_cls_name="AutoFeatureExtractor",
     )
 
     def load_model(self) -> torch.nn.Module:
--- a/neuralset-repo/neuralset/extractors/audio.py
+++ b/neuralset-repo/neuralset/extractors/audio.py
@@ -314,6 +314,11 @@ class SonarAudio(BaseAudio):
 
 class HuggingFaceAudioConfig(HuggingFaceConfig):
     processor_cls_name: str = "AutoFeatureExtractor"
+    HF_CLASS_DEFAULTS: tp.ClassVar[dict[str, dict[str, str]]] = {
+        "w2v-bert": {"model_cls_name": "Wav2Vec2BertModel"},
+        "m4t": {"model_cls_name": "SeamlessM4TModel"},
+        "whisper": {"model_cls_name": "WhisperModel"},
+    }
 
 
 class HuggingFaceAudio(BaseAudio, HuggingFaceMixin):
--- a/neuralset-repo/neuralset/extractors/audio.py
+++ b/neuralset-repo/neuralset/extractors/audio.py
@@ -389,7 +389,7 @@ class HuggingFaceAudio(BaseAudio, HuggingFaceMixin):
         from transformers import AutoFeatureExtractor
 
         return AutoFeatureExtractor.from_pretrained(
-            model_name, **self._hf_processor_kwargs()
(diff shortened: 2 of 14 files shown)


neuralset-repo/neuralset/events/transforms/test_transforms.py
```diff
@@ -1065,7 +1065,10 @@ def test_similarity_splitter(image_fpaths: list[Path]) -> None:
         }
     )
 
-    feat = ns.extractors.image.HuggingFaceImage(device="cpu", layers=1.0)
+    feat = ns.extractors.image.HuggingFaceImage(
+        hf_config={"device_map": "cpu"},  # type: ignore[arg-type]
```
kingjr: not a huge fan of the variable names

neuralset-repo/neuralset/extractors/audio.py
```diff
@@ -379,24 +379,18 @@ def feature_extractor(self) -> nn.Module:
             self._feature_extractor = self._get_feature_extractor(self.model_name)
         return self._feature_extractor
 
-    @property
-    def model(self) -> nn.Module:
-        if not hasattr(self, "_model"):
-            self._model = self._get_sound_model(self.model_name)
-        return self._model
+    def _load_cached_model(self) -> torch.nn.Module:
+        return self._get_sound_model(self.model_name)
```
kingjr: do we need these nested methods?

neuralset-repo/neuralset/extractors/audio.py
```diff
@@ -476,14 +470,9 @@ class Wav2VecBert(HuggingFaceAudio):
...
-
-    def _get_sound_model(self, model_name: str) -> torch.nn.Module:
-        from transformers import Wav2Vec2BertModel
-
-        _model = Wav2Vec2BertModel.from_pretrained(model_name)
-        _model.to(self.device)
-        _model.eval()
-        return _model
+    hf_config: HuggingFaceConfig = HuggingFaceConfig(
+        model_cls_name="Wav2Vec2BertModel",
```
kingjr, line 468: why not model_name like the rest of the api?

neuralset-repo/neuralset/extractors/audio.py
```diff
@@ -502,14 +491,12 @@ class SeamlessM4T(HuggingFaceAudio):
     """
 
     model_name: str = "facebook/hf-seamless-m4t-medium"
```
kingjr, line 487: Is there an argument to have this dissociation between model_name and hf_config? model_name could have been part of the hf_config OR hf_config at the root. This dissociation feels a bit clunky.

Maybe it's just me / a varname issue, e.g. model_infra?

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -479,63 +592,75 @@ class HuggingFaceMixin(base.BaseModel):
...
+        else:
+            config = AutoConfig.from_pretrained(
+                self.model_name,
+                **hf_config.config_kwargs,
+            )
+            model = Model.from_config(config)
+            if isinstance(torch_dtype, torch.dtype):
+                model.to(dtype=torch_dtype)
+            model.to(self.device)
+        model.eval()
```
kingjr, line 711: naive question: do we usually set eval by default?

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -438,6 +437,120 @@ def _skip_new_event_types(key, val, prev):
...
+            "float32",
+            "float64",
+            "bfloat16",
+        ]
+        | None
+    ) = None
+    attn_implementation: str | None = None
+    revision: str | None = None
+    trust_remote_code: bool = False
+    model_cls_name: str | None = None
```
kingjr: when can this be None? SHouldn't we set to AutoModel?

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -438,6 +437,120 @@ def _skip_new_event_types(key, val, prev):
...
+    processor_cls_name: str | None = None
+
+    @classmethod
+    def _exclude_from_cls_uid(cls) -> list[str]:
+        return ["device_map", "attn_implementation", "trust_remote_code"]
```
kingjr: are we sure attn_implementation doesn't affect the results?

(same question for device actually, I suspect there are edge cases but we'd like the device to not matter)

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -438,6 +437,120 @@ def _skip_new_event_types(key, val, prev):
...
+
+    def processor_cls(self) -> tp.Any:
+        return self._transformers_cls(
+            self.processor_cls_name,
+            "AutoProcessor",
+            "processor",
+        )
+
+    @property
+    def config_kwargs(self) -> dict[str, tp.Any]:
```
kingjr: .config? .to_dict()?

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -438,6 +437,120 @@ def _skip_new_event_types(key, val, prev):
...
+        return {
+            key: val
+            for key, val in {
+                "revision": self.revision,
+                "trust_remote_code": self.trust_remote_code or None,
+            }.items()
+            if val is not None
+        }
```
kingjr: I'm unclear about what s going on here

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -438,6 +437,120 @@ def _skip_new_event_types(key, val, prev):
...
+
+    @property
+    def torch_dtype_value(self) -> torch.dtype | tp.Literal["auto"] | None:
+        if self.torch_dtype is None or self.torch_dtype == "auto":
+            return self.torch_dtype
+        return getattr(torch, self.torch_dtype)
```
kingjr: shall we set this at model_post_init and store as private attr?

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -438,6 +437,120 @@ def _skip_new_event_types(key, val, prev):
...
+    @property
+    def model_kwargs(self) -> dict[str, tp.Any]:
+        kwargs = self.config_kwargs.copy()
+        if self.device_map == "auto":
+            kwargs["device_map"] = self.device_map
+        if self.attn_implementation is not None:
+            kwargs["attn_implementation"] = self.attn_implementation
+        if self.torch_dtype_value is not None:
+            kwargs["torch_dtype"] = self.torch_dtype_value
+        return kwargs
```
kingjr: I'm wondering whether we shouldn't instead use a private attr at the top of the class 
_CONFIG_VARS = ['revision', ..]
_MODEL_VARS = ['device_map', ...]

to avoid risking forgetting some elements in the future

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -479,63 +592,75 @@ class HuggingFaceMixin(base.BaseModel):
...
         if self.cache_n_layers is not None:
             excluded.extend(["layers", "layer_aggregation"])
         return excluded
 
+    @property
+    def device(self) -> str:
+        device_map = self.hf_config.device_map
+        if device_map == "auto":
+            return "cuda" if torch.cuda.is_available() else "cpu"
+        return device_map
```
kingjr: at model_post_init ? not sure @jrapin ?

neuralset-repo/neuralset/extractors/image.py
```diff
@@ -343,20 +261,35 @@ def model_post_init(self, log__):
...
+    def _full_predict(  # return the raw output, used in tests
+        self, images: torch.Tensor, text: str | list[str] = ""
+    ) -> tp.Any:
+        kwargs: dict[str, tp.Any] = dict(
+            images=[i.float() for i in images], return_tensors="pt"
+        )
+        if text:
+            kwargs["text"] = text
+        inputs = self.processor(**kwargs)
+        _fix_pixel_values(inputs)
```
kingjr, line 246: this feels wrong. are you sure?

kingjr:
You probably have the right design, but I'd love to have a live discussion with you and @jrapin  lucy to triple check

neuralset-repo/neuralset/events/transforms/test_transforms.py
sdascoli: Yeah it's a bit ugly I agree... I wanted model_config but protected by pydantic :(
If you have other ideas

neuralset-repo/neuralset/extractors/audio.py
sdascoli: Agreed, let's simplify. This was because technically load_model loads the whole model while _load_cached_model only takes a part, like model.speech_encoder for audio, but who cares

neuralset-repo/neuralset/extractors/audio.py
sdascoli: model_cls_name is used to specify the HF class when AutoModel doesn't work

neuralset-repo/neuralset/extractors/audio.py
sdascoli: Here's my rationale:

Why model_name not in hf_config:
- To keep previous API intact
- Because most of the time you would leave the default hf_config, and it's easier to write `HuggingFaceText(model_name="gpt2")` than `HuggingFaceText(hf_config=dict(model_name="gpt2"))`

Why hf_config not in root: 
- Because I think it's quite self-contained (all the arguments which are passed to AutoModel.from_pretrained()
- Because it would incur many arguments at the root, most of these arguments are only used in rare situations like for Sophia's benchmark or for non-standard models, so nice to hide them in an "advanced" subfield

neuralset-repo/neuralset/extractors/base.py
sdascoli: I believe so yes

neuralset-repo/neuralset/extractors/base.py
sdascoli: Yes maybe that would be clearer you're right

neuralset-repo/neuralset/extractors/base.py
sdascoli: In principle yes: this is to choose sdpa, flash_attention etc which are functionally equivalent, not "linear" variants of attention

neuralset-repo/neuralset/extractors/base.py
sdascoli: We need to disambiguate config_kwargs and model_kwargs. Renaming to config_build_kwargs and model_build_kwargs for clarity

neuralset-repo/neuralset/extractors/base.py
sdascoli: Removing this method, only used once

neuralset-repo/neuralset/extractors/base.py
sdascoli: What would be the advantage? 
I'm not a fan of privatte attrs, and think it's better to remain stateless:
- They're ugly and not easy to understand for newcomers
- I have had issues in the past with private attrs, eg if you override model_post_init in a subclass and forget to call super().model_post_init(), or if you reinstantiate a class with model_dump

neuralset-repo/neuralset/extractors/base.py
sdascoli: Same as above, I prefer properties than private attrs for stuff which is instantaneous to compute

neuralset-repo/neuralset/extractors/image.py
sdascoli: This is copied from previous behaviour, I think it's to add caption

sdascoli:
> You probably have the right design, but I'd love to have a live discussion with you and @jrapin lucy to triple check

Thanks for the quick and thorough feedback, just pushed some changes to address it!

neuralset-repo/neuralset/extractors/base.py
kingjr: really minor:
it would be a private attribute for the var names, not variables themselves. I think it's functionally equivalent, but allows it to be defined right below the class params. (i.e. if we had new config/model params, we'll more likely see that we need to update those fields and/or we can add a model_post_init to check that all variables are either in model or in config kwargs)

neuralset-repo/neuralset/extractors/base.py
kingjr: scratch that, if we submit from a non-cuda node, we don't want to freeze the attr

neuralset-repo/neuralset/extractors/image.py
kingjr: ok, let's deal with this later on. can you add a # FIXME

neuralset-repo/neuralset/extractors/base.py
```diff
@@ -479,63 +582,72 @@ class HuggingFaceMixin(base.BaseModel):
...
-                self._REPOS.append(name)
-                self._REPOS.sort()
-                try:
-                    fp.write_text("\n".join(self._REPOS))
-                except Exception:
-                    pass  # nevermind if there is no write permission
-                return True
-            except Exception:
-                return False
+        self.hf_config.check_model_instantiates(self.model_name)
```
jrapin: i'd expect this is way too heavy for the model post init, is it really instantiating a model?? if yes, do remove, this is way too huge a cost for initializing a config

neuralset-repo/neuralset/extractors/video.py
```diff
@@ -92,7 +89,6 @@ class HuggingFaceVideo(extractor_base.BaseExtractor, extractor_base.HuggingFaceM
         "julius>=0.2.7",
     )
     model_name: str = "MCG-NJU/videomae-base"
-    pretrained: bool = True
```
jrapin, line 95: why is this removed?

neuralset-repo/neuralset/extractors/video.py
```diff
@@ -134,6 +130,15 @@ def _validate_model_name(cls, model_name: str) -> str:
 
     def model_post_init(self, log__: tp.Any) -> None:
         super().model_post_init(log__)
+        if (
+            "vjepa2" in self.model_name.lower()
+            and self.hf_config.processor_cls_name != "AutoVideoProcessor"
+        ):
+            msg = (
+                "V-JEPA2 models require "
+                "hf_config.processor_cls_name='AutoVideoProcessor'."
```
jrapin: is it specific to video? + i'd expect this value should be the default?

neuralset-repo/neuralset/extractors/video.py
sdascoli: Moved to HuggingFaceMixin

neuralset-repo/neuralset/extractors/video.py
sdascoli: Looks like VideoMAE doesn't use it :/ But maybe I should put model_name="vjepa2" and processor_cls_name="AutoVideoProcessor" indeed

neuralset-repo/neuralset/extractors/base.py
sdascoli: Changed, thanks

jrapin (approved):

kingjr:
🚀🧹🧽🧼 thanks @sdascoli !

PR #138, opened 2026-06-09
Adjust duration calculation to account for precision

julien-gadonneix:
## What does this PR do?

Small fix for decimal precision in `EnsureTexts` `EventsTransform` to be sure that all the `Word` events are enclosed in the newly created `Text` events.
Reused the same approach as in `_extract_sentences` in `utils`

### DIFF
--- a/neuralset-repo/neuralset/events/transforms/splitting.py
+++ b/neuralset-repo/neuralset/events/transforms/splitting.py
@@ -157,7 +157,7 @@ class AssignWordSplitAndContext(EventsTransform):
     ratios: tuple[float, float, float] = (0.8, 0.1, 0.1)
     seed: int = 0
     max_unmatched_ratio: float = 0.0
-    tol: float = 0.0
+    tolerance: float = 0.0
     sentence_only: bool = True
     max_context_len: int | None = None
     override_sentences: bool = False
--- a/neuralset-repo/neuralset/events/transforms/splitting.py
+++ b/neuralset-repo/neuralset/events/transforms/splitting.py
@@ -157,6 +157,7 @@ class AssignWordSplitAndContext(EventsTransform):
     ratios: tuple[float, float, float] = (0.8, 0.1, 0.1)
     seed: int = 0
     max_unmatched_ratio: float = 0.0
+    tol: float = 0.0
     sentence_only: bool = True
     max_context_len: int | None = None
     override_sentences: bool = False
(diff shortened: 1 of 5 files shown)


kingjr:
I'm not entirely sure this is the right place for the fix, I'll defer to @jrapin

kingjr:
can you add a unit test?

kingjr:
@jrapin ok for you?

bkowshik:
Following up on @kingjr's question about whether this is the right place for the fix — I think it might be worth handling it one layer down. The root cause is a comparison issue rather than a data one. 

One side effect: there are now two workarounds for the same issue — this PR pads `duration` by `+ 2*eps`, while `_extract_sentences` in `utils.py` pads both edges (`start - eps` and `+ 2*eps`).

An alternative would be a single tolerance in `find_enclosed` (only two real callers):

```python
def find_enclosed(df, start, duration, *, tolerance: float = 1e-6):
    estart = np.asarray(df.start, dtype=float)
    estop = estart + np.asarray(df.duration, dtype=float)
    is_enclosed = (estart >= start - tolerance) & (estop <= start + duration + tolerance)
    return pd.Series(df.index[is_enclosed])
```

`EnsureTexts` could then revert to `duration = stop - start` (and the `eps` in `_extract_sentences` could go too) — same `1e-6`, just moved to where the rounding is compared, so durations stay accurate and the edge case is handled in one place. The existing test passes as-is since it already asserts via `find_enclosed`.

Either approach fixes the immediate bug — flagging the option in case it's useful.

kingjr:
definitely better to add tolerance to find_enclosed.

But I **think** (not sure) it should be 0 by default

kingjr (dismissed):
add tolerance to find_enclosed.

neuralset-repo/neuralset/events/transforms/splitting.py
```diff
@@ -157,6 +157,7 @@ class AssignWordSplitAndContext(EventsTransform):
     ratios: tuple[float, float, float] = (0.8, 0.1, 0.1)
     seed: int = 0
     max_unmatched_ratio: float = 0.0
+    tol: float = 0.0
```
kingjr: tolerance

neuralset-repo/neuralset/events/transforms/text.py
```diff
@@ -133,11 +133,14 @@ class AddSentenceToWords(EventsTransform):
     ----------
     max_unmatched_ratio : float
         Maximum ratio of words without a character-positioned match.
+    tol : float, optional
+        Tolerance for Text event start and end times. Default is 0.0.
```
kingjr: 0. sec.

kingjr:
thanks @julien-gadonneix !

PR #150, opened 2026-06-12
fix neuralfetch quickstart: use registered study and quickstart extra

### LINKED ISSUE #149: NeuralFetch landing-page quickstart is broken: unregistered study `Gwilliams2022Neural` + missing download-backend extra
### Description

The quickstart on the [NeuralFetch landing page](https://facebookresearch.github.io/neuroai/neuralfetch/index.html) (`docs/neuralfetch/index.rst`) cannot run as written. It fails for **two** independent reasons:

1. It calls `ns.Study(name="Gwilliams2022Neural", ...)`, but `Gwilliams2022Neural` is registered in **no** package, so it raises `ImportError`.
2. Its install line `pip install neuralfetch` does not pull any download backend, so even a *valid* study's `download()` raises `ModuleNotFoundError: pip install openneuro-py`. The backends live in the `neuralfetch[quickstart]` extra, which the page never mentions.

A first-time user copy-pasting the quickstart hits (1) immediately, and (2) right after.

### Steps to reproduce

**Bug 1 — `Gwilliams2022Neural` is not a registered study**

```python
import neuralset as ns
ns.Study(name="Gwilliams2022Neural", path="/data")
```

```python
ImportError: Study 'Gwilliams2022Neural' not found
(scanned: neuralset.events.testing, neuralfetch),
import the module that defines it before use.
```

**Bug 2 — quickstart install omits the download backend**

Even with a valid study, the documented install is insufficient:

```python
study = ns.Study(name="Bel2026PetitListenSample", path="/data")
study.download()
```

```python
ModuleNotFoundError: Bel2026PetitListenSample requires packages that are not installed.
  pip install openneuro-py
```

`openneuro-py` (and the other backends) are declared in `neuralfetch[quickstart]`, not in the base install. The page's note "no extra imports needed" is true for *imports* but misleading about *installs*.

### Environment

Latest PyPI releases, reproduced on Google Colab.

| package | version |
|---|---|
| python | 3.12.13 |
| neuralset | 0.2.2 |
| neuralfetch | 0.2.2 |
| openneuro-py | 2026.4.1 |
| mne | 1.12.1 |
| pandas | 3.0.3 |

bkowshik:
## What does this PR do?

The landing-page quickstart could not run as written:

- `Gwilliams2022Neural` is not a registered study (absent from `Study.catalog()`), so `ns.Study(name=...)` raised `ImportError`.
- `pip install neuralfetch` pulls no download backend, so `download()` raised `ModuleNotFoundError: pip install openneuro-py`.

Switch the example to `Bel2026PetitListenSample` (MEG + spoken Le Petit Prince, OpenNeuro ds007523) and add the `pip install 'neuralfetch[quickstart]'` line, mirroring the bare-then-extras install shown on the neuralset and neuraltrain landing pages. Drop the hand-written output block (no other landing page pastes command output) and wrap the final call in `print(...)`.

## Related Issues

Closes #149

## Checklist

- [ ] Tests added or updated
- [ ] Docstrings updated for any changed public APIs

### DIFF
--- a/docs/neuralfetch/index.rst
+++ b/docs/neuralfetch/index.rst
@@ -44,25 +44,22 @@ Quick install
    pip install neuralfetch
 
 Installing NeuralFetch automatically registers all curated studies in
-NeuralSet's catalog — no extra imports needed.
+NeuralSet's catalog — no extra imports needed. The public-data studies
+download through pluggable backends (OpenNeuro, OSF, S3, …) that ship
+with the ``quickstart`` extra:
+
+.. code-block:: bash
+
+   pip install 'neuralfetch[quickstart]'
 
 .. code-block:: python
 
    import neuralset as ns
 
-   study = ns.Study(name="Gwilliams2022Neural", path="/data")  # MEG + speech, from OSF
+   study = ns.Study(name="Bel2026PetitListenSample", path="/data")  # MEG + speech, OpenNeuro ds007523
    study.download()
    events = study.run()
-   events[["type", "start", "duration", "subject", "text"]].head()
-
-.. code-block:: text
-
-   type   start  duration              subject          text
-   Meg      0.0     396.0  Gwilliams2022Neural/A0001           NaN
-   Audio    0.0      42.3  Gwilliams2022Neural/A0001           NaN
-   Word     1.52     0.22  Gwilliams2022Neural/A0001         there
-   Word     1.74     0.18  Gwilliams2022Neural/A0001           was
-   Word     1.92     0.08  Gwilliams2022Neural/A0001             a
+   print(events[["type", "start", "duration", "subject", "text"]].head())
 
 ----
 
--- a/docs/neuralfetch/index.rst
+++ b/docs/neuralfetch/index.rst
@@ -44,9 +44,8 @@ Quick install
    pip install neuralfetch
 
 Installing NeuralFetch automatically registers all curated studies in
-NeuralSet's catalog — no extra imports needed. The public-data studies
-download through pluggable backends (OpenNeuro, OSF, S3, …) that ship
-with the ``quickstart`` extra:
+NeuralSet's catalog — no extra imports needed. Downloading their data
+needs the backends in the ``quickstart`` extra:
 
 .. code-block:: bash
 


kingjr:
@kbegany is this related to our study renaming issue?

bkowshik:
Reopening this — sorry, I closed it by mistake while cleaning up some old branches. Nothing's changed on the PR itself, so it's good to pick up where we left off.

Thanks for your patience!

docs/neuralfetch/index.rst
```diff
@@ -44,25 +44,22 @@ Quick install
    pip install neuralfetch
 
 Installing NeuralFetch automatically registers all curated studies in
-NeuralSet's catalog — no extra imports needed.
+NeuralSet's catalog — no extra imports needed. The public-data studies
```
kingjr: unncessary comment IMO

docs/neuralfetch/index.rst
```diff
@@ -44,25 +44,22 @@ Quick install
...
+.. code-block:: bash
+
+   pip install 'neuralfetch[quickstart]'
 
 .. code-block:: python
 
    import neuralset as ns
 
-   study = ns.Study(name="Gwilliams2022Neural", path="/data")  # MEG + speech, from OSF
+   study = ns.Study(name="Bel2026PetitListenSample", path="/data")  # MEG + speech, OpenNeuro ds007523
```
kingjr, line 58: I'd rather fix the root issue. Gwilliams should work too

docs/neuralfetch/index.rst
bkowshik: I've **trimmed the prose** per your other comment, keeping the `neuralfetch[quickstart]` install — that missing backend is an independent, confirmed bug (#149).

docs/neuralfetch/index.rst
bkowshik: `Gwilliams2022Neural` isn't a registered study in `neuralfetch` today — no class under `neuralfetch/studies/`, and the only repo references are this doc and a dead (overwritten) line in neuralset's `test_transforms.py`. The dataset is public (MEG-MASC — OSF `ag3kj` / OpenNeuro `ds004276`), so it's portable, just not shipped.

Happy to fix the root cause instead of swapping the example, but want to head the right way given @kbegany's renaming question:

- Was `Gwilliams2022Neural` **renamed** (if so, to what)? Or is it **net-new** and needs registering under `neuralfetch/studies/`?
- If net-new, in scope here, or a separate study-contribution PR with the doc pointing at a shipped study meanwhile?

kingjr: @kbegany @teonbrooks @hubertjb gwilliams2022 should be shipped, shouldn't it?

kingjr (changes requested):
ship Gwilliams2022

docs/neuralfetch/index.rst
teonbrooks: @kingjr, I just checked and it's not currently in this repo.

PR #157, opened 2026-06-15
Fix MNE sample-data re-downloads in Mne2013Sample/Fake2025Meg

### LINKED ISSUE #153: `Fake2025Meg` / `Mne2013Sample` re-download data on `run()` after `download()`
### Description

- **Package:** `neuralset`
- **File:** `neuralset/events/testing/mne2013sample.py`
- **Affected:** `Mne2013Sample`, `Mne2013SampleEeg`, `Fake2025Meg`

## Summary

`download()` and `run()` resolve the MNE sample-data directory to **three different roots**. Calling `download()` then `run()` (a natural sequence) makes `run()` look in empty directories and re-download the whole 1.65 GB dataset **twice** — ending with three copies on disk (~4.7 GB).

## Reproduction

```python
import neuralset as ns
study = ns.Study(name="Fake2025Meg", path=ns.CACHE_FOLDER)
study.download()
study.run()
```

```bash
find ~/.cache/neuralset -type d -name "MNE-sample-data"
```

## Observed (real run)

After `download()` — one copy:
```
/root/.cache/neuralset/Fake2025Meg/download/MNE-sample-data
```

`run()` then downloads the dataset two more times:
```
Downloading file 'MNE-sample-data-processed.tar.gz' ... to '/root/.cache/neuralset/Fake2025Meg'.
100%|...| 1.65G/1.65G
...
Downloading file 'MNE-sample-data-processed.tar.gz' ... to '/root/.cache/neuralset'.
100%|...| 1.65G/1.65G
```

After `run()` — three copies:
```
/root/.cache/neuralset/MNE-sample-data
/root/.cache/neuralset/Fake2025Meg/MNE-sample-data
/root/.cache/neuralset/Fake2025Meg/download/MNE-sample-data
```

## Root cause

Three methods disagree on the MNE `data_path` root (with `path=ns.CACHE_FOLDER`):

| Method | Code | Resolves to |
|---|---|---|
| `_download()` (writer) | [`mne2013sample.py:62`](neuralset/events/testing/mne2013sample.py#L62) — `self.path / "download"` | `…/Fake2025Meg/download/MNE-sample-data` |
| `iter_timelines()` | [`mne2013sample.py:170`](neuralset/events/testing/mne2013sample.py#L170) — `data_path(self.path)` | `…/Fake2025Meg/MNE-sample-data` |
| `_get_data_path()` (reader) | [`mne2013sample.py:71`](neuralset/events/testing/mne2013sample.py#L71) — `data_path(ns.CACHE_FOLDER)` | `…/MNE-sample-data` |

The divergence is **order-dependent**: `Study.download()` force-appends `/<ClassName>` to `self.path` ([`study.py:394`](neuralset/events/study.py#L394)), so afterwards the three methods point at three different roots. MNE only skips a download when `<path>/MNE-sample-data` already exists (and it discards the archive after untar), so each mismatched root triggers a fresh download.

The documented `run()`-only flow (no prior `download()`) is unaffected: `self.path` stays `= CACHE_FOLDER`, all roots collapse to `…/MNE-sample-data`, and only one copy is fetched.

## Expected vs actual

- **Expected:** `run()` after `download()` reuses the downloaded data — no re-download.
- **Actual:** `run()` re-downloads MNE-sample-data twice into two other directories.

### Steps to reproduce

```python
import neuralset as ns
study = ns.Study(name="Fake2025Meg", path=ns.CACHE_FOLDER)
study.download()
study.run()
```

### Environment

- OS: Linux-6.6.122+-x86_64-with-glibc2.35
- Python version: 3.12.13
- neuralset version: 0.2.2

bkowshik:
Fixes #153

---

`Mne2013Sample`/`Fake2025Meg` resolved the MNE sample dataset through three different roots, so `download()` then `run()` re-downloaded the 1.65 GB dataset twice (3 copies, ~4.7 GB). Two related bugs on the same paths are fixed too:

- `run()` showed **no download progress bar** (`verbose=True` was only in `_download()`).
- `run()` then `download()` crashed with `RuntimeError: ... instance was frozen`.

## Changes

- **`mne2013sample.py`**: funnel `_download`, `iter_timelines`, `_get_data_path`, and `_load_timeline_events` through one `_download_root()`/`_mne_data_path()` helper (anchored on `<self.path>/download`, `verbose=True`). Drops the `CACHE_FOLDER` special-case and fallback.
- **`study.py`**: `download()` now tolerates a frozen instance (after `run()`) instead of crashing — `path` is excluded from the cache UID, so it's safe. Additive; download-first flow unchanged.
- **Tests**: mocked single-root/progress-bar test (no network), plus a frozen-instance `run()`→`download()` test.

Result: every order yields one copy, shows the progress bar, and never crashes.

## Test

```bash
pytest neuralset/events/testing/test_mne2013sample.py neuralset/events/test_study.py::test_download_after_run_does_not_crash_when_frozen
```

## What I learned

Tracing this bug taught me a lot about how `neuralset` and `neuralfetch` are wired — the `Study` download/path lifecycle, the per-study `self.path`/`download/` convention shared across studies, and how `exca` freezes a model after `run()`. That context is what made the root cause and the right-sized fix clear.

### DIFF
--- a/neuralfetch-repo/neuralfetch/test_sivakumar2024emg2qwerty.py
+++ b/neuralfetch-repo/neuralfetch/test_sivakumar2024emg2qwerty.py
@@ -44,16 +44,20 @@ def _make_bids_tree(root, subdir="download"):
 
 @pytest.fixture
 def bids_tree(tmp_path):
-    sub, ses, _ = _make_bids_tree(tmp_path)
-    return tmp_path, sub, ses
+    # build under the study's resolved path so it matches ``bids_root``
+    study = Sivakumar2024Emg2qwerty(path=str(tmp_path))
+    sub, ses, _ = _make_bids_tree(study.path)
+    return study.path, sub, ses
 
 
 def test_emg2qwerty_study_source(tmp_path):
     """``iter_timelines`` / ``_load_timeline_events`` / ``bids_root`` work
     on a BIDS tree placed under ``download/`` (the layout
     ``Study.download`` produces)."""
-    sub, ses, bids_root = _make_bids_tree(tmp_path)
     study = Sivakumar2024Emg2qwerty(path=str(tmp_path))
+    # the study subfolder is resolved in model_post_init, so build the tree
+    # under the study's resolved path (``<tmp_path>/Sivakumar2024Emg2qwerty``).
+    sub, ses, bids_root = _make_bids_tree(study.path)
 
     assert study.bids_root == bids_root
     assert list(study.iter_timelines()) == [{"subject": sub, "session": ses}]
--- a/neuralset-repo/neuralset/events/study.py
+++ b/neuralset-repo/neuralset/events/study.py
@@ -382,22 +382,9 @@ class Study(base.Step):
     @tp.final
     def download(self, **kwargs: tp.Any) -> None:
         self._check_requirements()
-        # Ensure download goes into a subfolder named after the study
-        name = self.__class__.__name__
-        if self.path.name.lower() != name.lower():
-            try:
-                self.path = self.path / name
-                STUDY_PATHS[self.__class__.__name__] = self.path
-                logger.info("Download path updated to %s", self.path)
-            except RuntimeError as e:
-                # Instance frozen by a prior run(): keep the path run() already
-                # resolved and download into it. ``path`` is excluded from the
-                # cache UID (see _exclude_from_cls_uid), so identity is unaffected.
-                if "frozen" not in str(e).lower():
-                    raise
-                logger.info(
-                    "Instance frozen; downloading into existing path %s", self.path
-                )
+        # The study subfolder is settled in ``model_post_init`` (see there), so
+        # ``download()`` never reassigns ``self.path`` and is safe to call on a
+        # frozen post-``run()`` instance.
         self.path.mkdir(parents=True, exist_ok=True)
         self._download(**kwargs)
         if not self.path.exists():
@@ -448,7 +435,14 @@ class Study(base.Step):
                 "or pass name= to dispatch: Study(name='MyStudy2024', path=...)"
             )
         name = self.__class__.__name__
+        # Settle the study subfolder once, here at construction, before exca can
+        # freeze the instance: prefer an existing ``<path>/<name>`` (or lowercase)
+        # folder, otherwise append ``<name>``. ``download()`` then never reassigns
+        # ``self.path`` — which would crash on a frozen post-``run()`` instance
+        # (issue #153, run -> download order).
         self.path = _identify_study_subfolder(self.path, name)
+        if self.path.name.lower() != name.lower():
+            self.path = self.path / name
         STUDY_PATHS[self.__class__.__name__] = self.path  # record for path lookup
         # Auto-propagate cache folder and mode so users only need to set it once
         if self.infra is not None:
--- a/neuralset-repo/neuralset/events/study.py
+++ b/neuralset-repo/neuralset/events/study.py
@@ -385,9 +385,19 @@ class Study(base.Step):
         # Ensure download goes into a subfolder named after the study
         name = self.__class__.__name__
         if self.path.name.lower() != name.lower():
-            self.path = self.path / name
-            STUDY_PATHS[self.__class__.__name__] = self.path
-            logger.info("Download path updated to %s", self.path)
+            try:
+                self.path = self.path / name
+                STUDY_PATHS[self.__class__.__name__] = self.path
+                logger.info("Download path updated to %s", self.path)
+            except RuntimeError as e:
+                # Instance frozen by a prior run(): keep the path run() already
+                # resolved and download into it. ``path`` is excluded from the
+                # cache UID (see _exclude_from_cls_uid), so identity is unaffected.
+                if "frozen" not in str(e).lower():
+                    raise
+                logger.info(
+                    "Instance frozen; downloading into existing path %s", self.path
+                )
         self.path.mkdir(parents=True, exist_ok=True)
         self._download(**kwargs)
         if not self.path.exists():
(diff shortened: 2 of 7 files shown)


neuralset-repo/neuralset/events/study.py
```diff
@@ -385,9 +385,19 @@ def download(self, **kwargs: tp.Any) -> None:
         # Ensure download goes into a subfolder named after the study
         name = self.__class__.__name__
         if self.path.name.lower() != name.lower():
-            self.path = self.path / name
-            STUDY_PATHS[self.__class__.__name__] = self.path
-            logger.info("Download path updated to %s", self.path)
+            try:
```
kingjr: unnecessary changes

neuralset-repo/neuralset/events/testing/test_mne2013sample.py
```diff
@@ -0,0 +1,63 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
kingjr, line 1: Probably not critical, I'd remove to not overload the CI

neuralset-repo/neuralset/events/study.py
bkowshik: This guards a real crash: `run()` then `download()` raises `RuntimeError: … instance was frozen` — `run()` freezes the model (exca) and `download()` then reassigns `self.path`. Ex:

```python
import neuralset as ns

study = ns.Study(name="Fake2025Meg", path=ns.CACHE_FOLDER)
study.run()        # exca freezes the instance here
study.download()   # 💥 reassigns self.path on the frozen model
```

neuralset-repo/neuralset/events/testing/test_mne2013sample.py
bkowshik: These are fully mocked — they patch `mne.datasets.sample.data_path` and never download the `1.65 GB` sample dataset, so they run in milliseconds with no network/data cost to CI. They pin the specific [#153](https://github.com/facebookresearch/neuroai/issues/153) regression (all paths resolve to one root → a single copy), which the existing run-only MNE tests don't assert.

Let me know if you want me to remove it though.

kingjr: great, thanks for checking!

neuralset-repo/neuralset/events/study.py
kingjr: mmmh, actually we shouldn't have changed self.path. path.lower() should be probably applied before/at model_post_init , wdyt?

neuralset-repo/neuralset/events/testing/test_mne2013sample.py
bkowshik: _NOTE: Posting for future reference._

## Test does not download the 1.65 GB MNE sample dataset

The test mocks the only downloader call site (`monkeypatch.setattr(mne.datasets.sample, "data_path", _fake)`), so no network/bytes. Confirmed empirically:

```bash
cd neuralset-repo
du -sh ~/mne_data 2>/dev/null                                               # before
time python -m pytest neuralset/events/testing/test_mne2013sample.py --basetemp=/tmp/t153
du -sh ~/mne_data 2>/dev/null                                               # after — unchanged
find /tmp/t153 -type f ; du -sh /tmp/t153                                   # no files; ~0 bytes
```

Result: `2 passed in 0.16s` · `~/mne_data` 63M → 63M · `find` empty · `/tmp/t153` = `0B`. A real fetch would take minutes and add 1.65 GB — the 0.16s runtime alone proves the real `data_path` was never hit. **No download.**

neuralset-repo/neuralset/events/study.py
bkowshik: Good call — done in 7419916. Moved the subfolder resolution into model_post_init so self.path is settled at construction; download() no longer touches it and the try/except is gone. This matches _BaseMoabb, which already remaps its path in model_post_init, and makes run()/download() agree on the path for every study, not just MNE. (One neuralfetch path test needed updating since the subfolder is now present at construction.)

neuralfetch-repo/neuralfetch/test_sivakumar2024emg2qwerty.py
```diff
@@ -44,16 +44,20 @@ def _make_bids_tree(root, subdir="download"):
 
 @pytest.fixture
 def bids_tree(tmp_path):
-    sub, ses, _ = _make_bids_tree(tmp_path)
-    return tmp_path, sub, ses
+    # build under the study's resolved path so it matches ``bids_root``
```
kingjr, line 47: it this related to mne sample?

kingjr (changes requested):
ship Gwilliams2022

neuralfetch-repo/neuralfetch/test_sivakumar2024emg2qwerty.py
bkowshik: It's a side-effect of moving subfolder resolution into `model_post_init` (per your earlier suggestion). That now resolves `<path>/<StudyName>` at construction for all studies, so this test (and the Zyma test in `test_studies.py`) had to build their BIDS trees under the resolved `<tmp_path>/Sivakumar2024Emg2qwerty` path instead of `<tmp_path>` directly. No behavior change to the study — just keeping the path assertions correct after the refactor.

bkowshik:
> ship Gwilliams2022

I think this review may have been intended for [#150](https://github.com/facebookresearch/neuroai/pull/150) — this PR is scoped to the MNE sample re-download fix ([#153](https://github.com/bkowshik/neuroai/issues/153)) and doesn't touch Gwilliams2022 (or any study registration). Could you confirm whether anything's actually blocking here? Happy to address the Gwilliams2022 shipping discussion over on [#150](https://github.com/facebookresearch/neuroai/pull/150).

kingjr (approved):

kingjr:
thanks @bkowshik !

PR #158, opened 2026-06-15
Add test case to verify class MneTimedArray and EegExtractor load the data with correct order

nguoithichkhampha:
## What does this PR do?
The current test case in `test_neuro.py` just verified the size and data shape. I add this test to verify the channel value, make sure we load the data in correct order


kingjr:
Thanks, If there's not code modification but 1 new test, I'm not sure this is necessary?

nguoithichkhampha:
my cowoker has raised the issue neuralset load the input data in incorrect order. just add this test to help community easy to verify if they have same concern.

kingjr:
maybe I'm missing something but this seems to only modify the tests, and thus does fix anything?

nguoithichkhampha:
yes, as in the previous comment. I explained the reason why I add new test case into `test_neuro`. If you think the change is not useful, you can close the PR.

PR #148, opened 2026-06-12
Add load_brainmarks_split utility and tests

kbegany:
## Summary

- Adds `load_brainmarks_split(dataset, subject_col)` to `neuralset/utils.py`:
  reads HuggingFace-style Arrow shards from
  `$NEURALSET_STUDY_FOLDER/Brainmarks/fmri-datasets/eval/<dataset>.arrow/`
  and returns a `{subject_id: split}` mapping
- Gracefully returns `{}` with a warning if `NEURALSET_STUDY_FOLDER` is
  unset or the dataset directory does not exist
- Adds three tests in `test_utils.py` covering the path, missing
  env var, and missing dataset directory

## Checklist

- [x] Tests added or updated: `pytest neuralset-repo/neuralset/test_utils.py -v -k brainmarks`
- [x] Docstrings updated for any changed public APIs
- [x] `pytest` and `mypy` pass locally

PR #159, opened 2026-06-16
chore: hoist core stdlib imports to module level

bkowshik:
Ref: https://github.com/facebookresearch/neuroai/pull/148#discussion_r3405652116

---

Moves a handful of stdlib imports (`os`, `shutil`, `tempfile`, `zipfile`, `importlib`) from inside function bodies up to module level, following the review note that core dependencies belong at the top. These are stdlib modules with no import cost, side effects, or circular-import risk, so deferring them bought nothing. In `download.py`, `tempfile`/`zipfile` were already imported at the top — the inline versions were just redundant re-imports.

Deliberate lazy imports (heavy/optional deps like torch and plotly, circular-import breakers, and the debug-only `pdb` import) are left as-is.

### DIFF
--- a/neuralbench-repo/neuralbench/cli.py
+++ b/neuralbench-repo/neuralbench/cli.py
@@ -14,6 +14,7 @@ config assembly are delegated to :mod:`neuralbench.registry` and
 
 import argparse
 import logging
+import os
 import sys
 import traceback
 import typing as tp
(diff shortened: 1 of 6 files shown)


kingjr:
LGTM, thanks @bkowshik !

PR #160, opened 2026-06-16
Remove duplicate _get_max_length method in HuggingFaceText

bkowshik:
## What

Removes a duplicated `_get_max_length` definition in `HuggingFaceText` (`neuralset-repo/neuralset/extractors/text.py`). The method was defined twice with byte-identical bodies.

## Why

Python keeps only the second definition, so the first was dead code. It slipped in via PR #147 and isn't caught by linting: ruff (F811) and pylint (E0102) both skip redefinitions of underscore-prefixed names by default.

## Change

- Deleted the first of the two identical definitions (18 lines).
- No behavior change: the surviving method is identical, and the sole call site (`_get_data`) is untouched.

### DIFF
--- a/neuralset-repo/neuralset/extractors/text.py
+++ b/neuralset-repo/neuralset/extractors/text.py
@@ -364,24 +364,6 @@ class HuggingFaceText(BaseStatic, HuggingFaceMixin):
                 return self._max_length
         return None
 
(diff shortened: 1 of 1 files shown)


kingjr:
LGTM, thanks @bkowshik !

PR #164, opened 2026-06-19
Honor FmriCleaner.ensure_finite when other cleaning options are disabled

### LINKED ISSUE #130: neuralset: FmriCleaner skips ensure_finite when other cleaning options are disabled
## Summary

`neuralset.extractors.neuro.FmriCleaner.clean()` bypasses `nilearn.signal.clean()` when `detrend=False`, `standardize=False`, no filters are configured, and no confounds are passed. Because `ensure_finite=True` is only forwarded inside that guarded call, disabling detrending/standardization also disables finite-value handling, allowing NaNs/Infs to leak into downstream models.

## Current behavior

The cleaner only calls `nilearn.signal.clean()` under this condition:

```python
if (
    self.detrend
    or self.standardize
    or (self.high_pass is not None)
    or (self.low_pass is not None)
    or (confounds is not None)
):
    data = nilearn.signal.clean(..., ensure_finite=self.ensure_finite, ...)
```

## Expected behavior

`ensure_finite=True` should still be honored when all other cleaning operations are disabled. Either:

- include `self.ensure_finite` in the guard that decides whether to call `nilearn.signal.clean()`, or
- handle finite replacement explicitly before returning when `ensure_finite=True`.

bkowshik:
## Summary

Fixes #130. `FmriCleaner.clean()` only applied `ensure_finite` inside the `nilearn.signal.clean()` call, which was skipped when `detrend`, `standardize`, and filters were all off. So `NaN`/`Inf` leaked through despite `ensure_finite=True`, silently corrupting downstream models.

## Change

- Add `ensure_finite` to the guard so it's honored independently. Routing through nilearn keeps the logic in one place.
- Fix the docstring (it also zeroes `±Inf`).
- Add a test covering both on/off behavior.

Defaults are unaffected. `ruff` clean; new test fails without the fix, passes with it.

### DIFF
--- a/neuralset-repo/neuralset/extractors/neuro.py
+++ b/neuralset-repo/neuralset/extractors/neuro.py
@@ -1074,7 +1074,7 @@ class FmriCleaner(pydantic.BaseModel):
     filter: tp.Literal["butterworth", "cosine"] | None
         Filter to use: "butterworth" or "cosine".
     ensure_finite: bool
-        Whether to set nans to 0.
+        Whether to replace non-finite values (NaN, +/-Inf) with 0.
     """
 
     model_config = pydantic.ConfigDict(extra="forbid")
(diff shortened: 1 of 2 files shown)


kingjr:
great catch, thanks @bkowshik !

PR #175, opened 2026-06-24
Fix Brennan2019Hierarchical loading for the v2 (`bn999738r`) release

### LINKED ISSUE #172: Crash in Brennan2019 loader: hardcoded trial count (len(meta) != 2129)
### Description

Getting a RuntimeError when loading the Brennan2019Hierarchical dataset. The _read_meta method crashes on subjects with missing segments because of this hardcoded check:
Python
        if len(meta) != 2129:
            raise RuntimeError(f"Expected 2129 trials, got {len(meta)}")  # FIXME

Also the next step (events = meta.join(story)) joins on row indices. If a subject is missing the first N trials, their remaining EEG data gets misaligned with the beginning of the story text.
How did you guys handle this for the benchmark results in the paper? Were these subjects dropped upstream, or is there a missing alignment step?

### Steps to reproduce

```python
Just instantiate the study and try to load the events. It crashes as soon as it hits an incomplete subject...
```

### Environment

- OS:
- Python version:
- Relevant package versions:

bkowshik:
Fixes #172

`Brennan2019Hierarchical` targeted the v2 Deep Blue record but the reader still assumed the v1 layout and channel naming, so `study.run()` failed on a fresh download. #172 reports the `len(meta) != 2129` crash — it's the deepest of four migration breakages, three of which block loading before you reach it. Fixes:

1. **`_download`** now extracts `proc.zip` (the `timelock-preprocessing/*.mat` files) in addition to `audio.zip`, skipping `__MACOSX` entries. The extract-success marker is bumped to `success_extract_v2` so existing caches re-extract — they'd otherwise keep the old marker and never pick up the `.mat` files.
2. **`iter_timelines`** now excludes the 6 in-range subjects with no `.mat` (`S28, S29, S31, S33, S46, S47`) alongside the 9 incomplete-trial subjects (1973–1987 trials vs 2129 — the existing `# FIXME … bad trl?` set), yielding the 33 already declared by `_info`/docstring.
3. **`_load_raw`** types any channel absent from the `easycap-M10` montage (`EOG`→eog, else misc) before `set_montage`, handling v2's `AUD` (a few subjects keep `Aux5`). The channel assertion changes from "62 total" to "60 scalp EEG" (matches `data_shape`), since the aux name/count varies per subject.

This resolves #172's crash by excluding the incomplete subjects (the drop path — matching the paper's 33-subject cohort), rather than recovering them via a fixed alignment for the `meta.join(story)` index join. That recovery is a separate, larger decision — see the #172 discussion.

Adds a dataset-free regression test pinning `len(iter_timelines()) == 33`.

## Verification

`study.run()` loads all 33 subjects; `compute_study_info` matches `_info` exactly (33 timelines/subjects, 2226 events, `data_shape=(60, 366525)`, 500 Hz). `ruff check`/`ruff format` and `pytest neuralfetch/test_studies.py` are green locally.

## Notes for review

I worked through this independently from the tutorials — debugging down the stack to all four breakages — and only found #172 and @hubertjb's comment when raising this PR. @hubertjb, since you flagged taking it this week, it's entirely yours to take over, modify, or close; I'm opening it in case it saves the time.

### DIFF
--- a/neuralfetch-repo/neuralfetch/studies/brennan2019hierarchical.py
+++ b/neuralfetch-repo/neuralfetch/studies/brennan2019hierarchical.py
@@ -98,33 +98,56 @@ class Brennan2019Hierarchical(study.Study):
             collection_id="cc387c09-b0e5-422b-a384-0d96e7ffdc73",
         ).download()
 
-        with success_writer(dl_dir / "success_extract") as already_done:
+        # Bumped from "success_extract": the v1 marker predates proc.zip
+        # extraction, so existing caches must re-extract to get the .mat files.
+        with success_writer(dl_dir / "success_extract_v2") as already_done:
             if not already_done:
-                print(f"Extracting `brennan2019` audio to {dl_dir}/audio...")
-                with zipfile.ZipFile(str(dl_dir / "audio.zip"), "r") as zip_:
-                    zip_.extractall(str(dl_dir))
+                # audio.zip -> audio/*.wav, proc.zip -> timelock-preprocessing/*.mat
+                for archive in ("audio.zip", "proc.zip"):
+                    print(f"Extracting `brennan2019` {archive} to {dl_dir}...")
+                    with zipfile.ZipFile(str(dl_dir / archive), "r") as zip_:
+                        members = [
+                            m for m in zip_.namelist() if not m.startswith("__MACOSX")
+                        ]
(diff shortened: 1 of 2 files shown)


hubertjb:
Thanks for the contribution @bkowshik, looks good to me! I confirm I was able to run the post-download extraction and the local tests.

PR #181, opened 2026-06-27
Make OpenNeuro downloads block under a running event loop (Jupyter/Colab)

bkowshik:
## Problem

Downloading any OpenNeuro-backed study from a notebook (Jupyter/Colab) reports success but leaves the dataset only partially downloaded. `study.run()` then fails with `FileNotFoundError` (e.g. `participants.tsv`) or `Dataset ... is corrupted, expected N timelines but found M`.

## Root cause

Upstream in openneuro-py, surfaced through our backend. Its `download()` only blocks when no event loop is running:

```python
try:
    loop = asyncio.get_running_loop()
    loop.create_task(coroutine)   # running loop (Jupyter): scheduled, never awaited → returns immediately
except RuntimeError:
    asyncio.run(coroutine)        # no loop (script/CLI): blocks
```

Under a notebook's running loop it returns *before any file is written* (then prints `Finished downloading`). `Openneuro` trusts it to be synchronous, so the success marker is written and `study.download()` logs success over an essentially empty directory — and `study.run()` races the still-downloading tree. CI never caught it because the tests don't download and pytest has no running loop, so openneuro always takes the blocking branch there.

## Fix

Run `on.download` on a worker thread, which has no running loop and so takes openneuro's blocking branch; the call returns only once the download is actually complete — matching the synchronous contract every other backend already honors. Public API only; no private internals or `nest_asyncio`.

## Scope & testing

- Fixes all OpenNeuro studies (7 studies / 12 call sites). No change to script/CLI/cluster paths, which already block.
- New offline regression tests: download completes synchronously under a running loop, and worker-thread errors propagate to the caller. Full `test_download.py`: 28 passed, ruff clean.

### DIFF
--- a/neuralfetch-repo/neuralfetch/download.py
+++ b/neuralfetch-repo/neuralfetch/download.py
@@ -1108,14 +1108,29 @@ class Openneuro(BaseDownload):
     nworkers: int = 5
 
     def _download(self) -> None:
+        from concurrent.futures import ThreadPoolExecutor
+
         import openneuro as on
 
-        on.download(
-            dataset=self.study,
-            target_dir=self._dl_dir,
-            include=self.include,
-            max_concurrent_downloads=self.nworkers,
-        )
+        # openneuro-py's ``download`` only blocks when no event loop is running
+        # (it falls back to ``asyncio.run``). Under an already-running loop --
+        # e.g. inside Jupyter/Colab -- it schedules the transfer with
+        # ``loop.create_task`` and returns *before any file is written*, so the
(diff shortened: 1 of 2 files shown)


teonbrooks:
thanks for the PR @bkowshik. I'm wondering, does it make sense to have this patch here or should it be upstreamed to openneuro-py? 

if i'm understanding correctly, we set one worker on the outer loop of openneuro, but we would be able to still use the `max_concurrent_downloads`? or would that go to 1 instead?

PR #189, opened 2026-07-02
Support image events in HuggingFaceVideo

sdascoli:
## Summary
- Extends `HuggingFaceVideo` to accept `event_types="Image"` and encode each image as a static repeated-frame clip with `num_frames` frames.
- Reuses the existing native video model prediction, token aggregation, and layer aggregation path for both real video clips and static image clips.
- Adds image-aware cache keys and a focused non-network test that verifies repeated static clips are sent to the video model.

## Issue Addressed
`HuggingFaceVideo` previously only accepted `Video` events, so image stimuli could not be embedded with native HuggingFace video models. This change lets image events be passed through the same video model path by converting each image into a static clip.

## Usage Example
```python
extractor = neuralset.extractors.HuggingFaceVideo(
    event_types="Image",
    model_name="MCG-NJU/videomae-base",
    num_frames=16,
    device="cpu",
)
embedding = extractor(image_event, start=image_event.start, duration=image_event.duration)
```

## Test plan
- `../.venv/bin/python -m pytest neuralset/extractors/test_video.py::test_video_requirements neuralset/extractors/test_video.py::test_huggingface_video_static_image_clip`
- `../.venv/bin/python -m mypy neuralset/extractors/video.py` currently reaches an unrelated baseline error in `neuralset/segments.py:397`; no errors are reported in `video.py`.

### DIFF
--- a/neuralset-repo/neuralset/extractors/test_video.py
+++ b/neuralset-repo/neuralset/extractors/test_video.py
@@ -101,7 +101,7 @@ def test_huggingface_video_static_image_clip(tmp_path: Path) -> None:
     filepath = tmp_path / "image.png"
     frame = np.random.randint(0, 256, (64, 64, 3), dtype=np.uint8)
     PIL.Image.fromarray(frame).save(filepath)
-    event = etypes.Image(start=1.0, duration=0.5, filepath=filepath, timeline="foo")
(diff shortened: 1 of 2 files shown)


sdascoli:
@claude review

PR #205, opened 2026-07-27
Fix Shirazi2024 (HBN) reaction-time window to be stimulus-locked

### LINKED ISSUE #195: Shirazi2024 (HBN) Reaction Time returns the wrong EEG window (response-locked instead of stimulus-locked)
## Summary

The EEG `reaction_time` task epochs its prediction window relative to the
**button press (Keystroke)** rather than the **stimulus (contrast-change)
onset**. This contradicts the window definition given in the EEG Foundation 2025
Challenge paper (arXiv:2506.19141), which specifies:

> "The window period is 0.5 to 2.5 seconds after stimulus onset for prediction."

And on which the [NeuralBench blog](https://facebookresearch.github.io/neuroai/neuralbench/tasks/eeg/reaction_time.html) relies on:
> "This task corresponds to Challenge 1 (“Cross-Task Transfer Learning”) of the [EEG Challenge 2025](https://eeg2025.github.io/) [[Aristimunha2025]](https://facebookresearch.github.io/neuroai/neuralbench/tasks/eeg/reaction_time.html#aristimunha2025)."

As currently configured, the window is `[button_press + 0.5, button_press + 2.5]`
— i.e. it sits entirely *after* the response, not after the stimulus.

## Location

`neuralbench-repo/neuralbench/tasks/eeg/reaction_time/config.yaml`

Current values:
```yaml
  target:
    name: EventField
    event_types: Keystroke
    event_field: reaction_time
    aggregation: trigger
  trigger_event_type: Keystroke
  start: 0.5
  duration: 2.0
```

## Why this produces a response-locked window

- In the study loader (`neuralfetch-repo/neuralfetch/studies/shirazi2024hbn.py`),
  a `Keystroke` event's `start` is the **button-press onset**
  (`right/left_buttonPress`, ~lines 388-398).
- The stimulus (contrast change / go-signal) is a **separate** `Stimulus`
  event (the `left/right_target` rows), and reaction time is defined as
  `reaction_time = start_button - start_target` (~line 408).
- Segment creation adds the `start` offset to the **trigger's** start
  (`neuralset-repo/neuralset/segments.py`, single-window branch:
  `seg_starts = (trigger_df["start"] + start)`).

So with `trigger_event_type: Keystroke` and `start: 0.5`, the window becomes
`[button_press + 0.5, button_press + 2.5]` — response-locked, contradicting the
paper's stimulus-locked spec.

## Suggested fix

Trigger on the stimulus (contrast-change/target) event instead of the keystroke:

```yaml
  target:
    name: EventField
    event_types: Stimulus      # was: Keystroke
    event_field: reaction_time
    aggregation: trigger
  trigger_event_type: Stimulus  # was: Keystroke
  start: 0.5
  duration: 2.0
```

This yields `[contrast_change + 0.5, contrast_change + 2.5]`, matching the paper.

The label still resolves: the study loader already merges `reaction_time` onto
the target/`Stimulus` rows (`shirazi2024hbn.py` ~lines 421-426), and the
existing `filter_reaction_time` query
(`"(reaction_time >= 0.5) | type == 'Eeg'"`) already isolates the target
events (the `trial` and `feedback` Stimulus subtypes have no `reaction_time`
and are dropped by the `>= 0.5` filter). Because that filter applies in both
cases, the *set of included trials/subjects is unchanged* — only the window
position moves from response-locked to stimulus-locked.

## Notes / verification

- The config has a single commit in history ("Adding NeuralBench", #42); it has
  been `Keystroke`-triggered since introduction — never stimulus-locked.
- No existing issue/PR appears to track this.
- I verified the proposed `Stimulus` trigger runs end-to-end against the
  packaged neuralbench/neuralset code (segment creation + the
  `EventField(aggregation="trigger")` target reading `reaction_time` off a
  `Stimulus` trigger both succeed).

### Steps to reproduce

```python
Running NeuralBench with Shirazi2024 (HBN) Reaction Time.
```

### Environment

- OS:
- Python version:
- Relevant package versions:

guptaishaan:
## Summary

Fixes #195.

The EEG `reaction_time` task triggers its prediction window on `Keystroke`, whose `start` is the button-press onset. The realized window is therefore `[button_press + 0.5, button_press + 2.5]` — entirely *after* the response.

The EEG Foundation Challenge 2025 paper (arXiv:2506.19141), which this task's docs cite, specifies:

> "The window period is 0.5 to 2.5 seconds after stimulus onset for prediction."

This switches the trigger to the contrast-change `Stimulus` event. `target.event_types` has to move with it: `aggregation: trigger` raises when the trigger is not an instance of the extractor's `event_types` (`neuralset/extractors/base.py:308-313`), so the two keys are not independently changeable.

## Verification

Confirmed without HBN data:

- `Keystroke` events carry the button-press onset — `shirazi2024hbn.py:380-390`, built from `right/left_buttonPress` with `onset` renamed to `start`.
- The contrast change is a separate `Stimulus` event — `:357-370`, the `left/right_target` rows with `event_type="target"`.
- `reaction_time = response − target` — `:400`, `rt["start_button"] - rt["start_target"]`.
- `reaction_time` is merged onto the `Stimulus`/target rows — `:413-418` — so `EventField(aggregation="trigger")` can read the label off a `Stimulus` trigger.
- The window is trigger-relative — `neuralset/segments.py:420`, `seg_starts = trigger_df["start"] + start`.
- `trial` and `feedback` `Stimulus` subtypes carry no `reaction_time` and are dropped by the existing `(reaction_time >= 0.5)` filter, so this change does not pull them into the trigger set.

Driving the real `list_segments` with the shipped `start=0.5, duration=2.0` (stimulus at t=11.0, press at t=12.2):

```
trigger=Keystroke  window=[12.70, 14.70]   <- begins 0.5 s after the response
trigger=Stimulus   window=[11.50, 13.50]   <- 0.5-2.5 s after stimulus onset
EventField(event_types='Stimulus', aggregation='trigger') -> 1.200
```

`StudyInfo._info` is unaffected — it declares `num_events_in_query=1, event_types_in_query={"Eeg"}`, independent of the task's `trigger_event_type`. Checked rather than assumed.

**Not verified:** I have no `NEURALSET_STUDY_FOLDER` / HBN data locally, so there is **no end-to-end benchmark run** behind this PR. I have not confirmed trial or subject counts against real data. Everything above is static reading plus execution of the loader and segmenter on synthetic events. A reviewer with data access should close that gap.

## ⚠️ Two things reviewers should weigh

**1. The trigger sets are not identical.** `shirazi2024hbn.py:415` merges reaction time onto the target with `button_df.groupby("trial_num")[...].max()`, so the loader explicitly anticipates multiple button presses per trial (`_get_trial_times:302-304` carries the comment *"handles cases of multiple responses"*). For a trial with two presses that both clear `reaction_time >= 0.5`:

```
Keystroke triggers: n=4  rts=[0.6, 1.2, 1.5, 1.9]
Stimulus  triggers: n=3  rts=[1.2, 1.5, 1.9]
```

So `Keystroke` yields one sample per qualifying *button press*, `Stimulus` one per *trial*, and for multi-press trials the label becomes the `.max()` — the slowest response — where `:302-304` uses `.min()` for the same situation elsewhere. I could not measure how often this fires without the dataset. If you want, this is measurable with:

```python
ccd = events[events.type == "Keystroke"]
multi = ccd[ccd.reaction_time >= 0.5].groupby(["timeline", "trial_num"]).size()
print(f"trials with >=2 qualifying presses: {(multi > 1).sum()} / {len(multi)}")
```

If you'd rather preserve the existing sample population exactly, this needs a loader-side decision, and I've deliberately left the loader untouched.

**2. Published results become non-comparable.** The config has a single commit in history ("Adding NeuralBench", #42), so this task has been response-locked since introduction — every NeuralBench number ever produced for Shirazi2024 reaction-time used the wrong window. Would you like a leaderboard note, a task-version bump, or something else alongside this?

## A supporting detail

The existing filter threshold is `reaction_time >= 0.5` and the window starts at `+0.5`. That threshold is only meaningful under stimulus-locking, where it guarantees the response falls at or after the window start; under response-locking it is arbitrary. The task docs agree — *"predict where, within a 2-s EEG window, participants responded"* — which requires the response to be inside the window.

## Tests

`neuralbench/test_task_configs.py` adds three data-free guards (no downloaded study needed):

- the task triggers on `Stimulus` with the paper's `0.5 / 2.0` offsets;
- `target.event_types` matches `trigger_event_type`, the invariant `aggregation: trigger` requires;
- segments built with the shipped offsets start at stimulus onset + 0.5, with the response falling inside the window.

All three fail on the pre-fix config (`assert 12.7 == 11.5 ± 1.2e-05`).

`ruff check .` clean, `ruff format --check .` clean, `pytest -x -q` → 184 passed.

Thanks to @Roysegg for the precise report and the loader line references.

### DIFF
--- a/neuralbench-repo/neuralbench/tasks/eeg/reaction_time/config.yaml
+++ b/neuralbench-repo/neuralbench/tasks/eeg/reaction_time/config.yaml
@@ -24,10 +24,14 @@ data:
   target:
     =replace=: true
     name: EventField
-    event_types: Keystroke
+    event_types: Stimulus
     event_field: reaction_time
     aggregation: trigger
-  trigger_event_type: Keystroke
+  # Trigger on the contrast-change (`target`) Stimulus, not the Keystroke, so the
+  # window below is 0.5-2.5 s after *stimulus onset* per Aristimunha et al. 2025
+  # (arXiv:2506.19141). `filter_reaction_time` above keeps only the `target`
+  # Stimulus rows: the `trial` and `feedback` subtypes carry no `reaction_time`.
+  trigger_event_type: Stimulus
   start: 0.5
   duration: 2.0
   summary_columns: [release]
--- a/neuralbench-repo/neuralbench/test_task_configs.py
+++ b/neuralbench-repo/neuralbench/test_task_configs.py
@@ -1,0 +1,100 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression guards for shipped task ``config.yaml`` semantics.
+
+These are data-free: they read the committed YAML and exercise the
+segmenter primitives, so they run without any downloaded study.
+"""
+
+from __future__ import annotations
+
+import pandas as pd
+import pytest
+
+from neuralset.events import standardize_events
+from neuralset.segments import list_segments
+
+from .registry import _resolve_task_dir, load_yaml_config
+
+
+@pytest.fixture
+def reaction_time_config() -> dict:
+    config = load_yaml_config(
+        _resolve_task_dir("eeg", "reaction_time") / "config.yaml", safe=True
+    )
+    assert config is not None
+    return config["data"]
+
+
+def test_reaction_time_is_stimulus_locked(reaction_time_config: dict) -> None:
+    """The EEG reaction-time window is locked to stimulus onset, not the response.
+
+    Regression guard for #195. ``Keystroke.start`` is the button-press onset, so
+    triggering on it puts the window entirely *after* the response. The EEG
+    Foundation Challenge 2025 spec (arXiv:2506.19141) defines the window as
+    0.5-2.5 s after *stimulus* onset.
+    """
+    assert reaction_time_config["trigger_event_type"] == "Stimulus"
+    assert reaction_time_config["start"] == 0.5
+    assert reaction_time_config["duration"] == 2.0
+
+
+def test_reaction_time_target_matches_trigger(reaction_time_config: dict) -> None:
+    """``aggregation: trigger`` requires the target to accept the trigger event.
+
+    ``BaseExtractor._select_events`` raises when the trigger is not an instance
+    of the extractor's ``event_types``, so these two keys must move together.
+    """
+    target = reaction_time_config["target"]
+    assert target["aggregation"] == "trigger"
+    assert target["event_types"] == reaction_time_config["trigger_event_type"]
+
+
+def test_reaction_time_window_follows_stimulus_onset(
+    reaction_time_config: dict,
+) -> None:
+    """Segments built with the shipped offsets sit after the stimulus, not the press.
+
+    Mirrors one CCD trial: contrast change at t=11.0, button press at t=12.2
+    (``reaction_time`` 1.2). Only the ``target`` Stimulus carries a
+    ``reaction_time``, which is what ``filter_reaction_time`` selects on.
+    """
+    start = reaction_time_config["start"]
+    duration = reaction_time_config["duration"]
+    stimulus_onset, press_onset = 11.0, 12.2
+
+    events = standardize_events(
+        pd.DataFrame(
+            [
+                dict(
+                    type="Stimulus",
+                    start=stimulus_onset,
+                    duration=2.4,
+                    timeline="tl0",
+                    reaction_time=press_onset - stimulus_onset,
+                ),
+                dict(
+                    type="Keystroke",
+                    start=press_onset,
+                    duration=0.1,
+                    timeline="tl0",
+                    text="left_buttonPress",
+                    reaction_time=press_onset - stimulus_onset,
+                ),
+            ]
+        )
+    )
+    triggers = (events.type == reaction_time_config["trigger_event_type"]) & events[
+        "reaction_time"
+    ].notna()
+
+    segments = list_segments(events, triggers=triggers, start=start, duration=duration)
+
+    assert len(segments) == 1
+    assert segments[0].start == pytest.approx(stimulus_onset + start)
+    # The response falls inside the window -- the premise of the task.
+    assert segments[0].start < press_onset < segments[0].stop


hubertjb:
Thanks for finding this bug and proposing a fix @guptaishaan! I ran the new config on the full HBN data and this indeed yields similar nRMSE values as those from the competition's leaderboard.

As you point out, editing the config in place changes the numbers already published in the docs and the arXiv paper for this task, so they'd no longer match. Instead, I'd favour using the repo's existing convention for this, i.e. _-prefixed task dirs are excluded from the default suite, while still being runnable by name.

This would mean:
- Keep the current tasks/eeg/reaction_time/ as-is (published results stay reproducible)
- Move your proposed corrected config to tasks/eeg/_reaction_time/
- Add a "Known issue / upcoming fix" note in reaction_time.rst explaining the response- vs stimulus-locked bug (with the paper citation) and pointing to _reaction_time

We would then promote _reaction_time → reaction_time at the next NeuralBench version bump, and refresh the docs/leaderboard.

Let me know if you'd like to include this in your PR, or if we should pick it up.

Roysegg:
Hi!

Thanks for addressing the bug @guptaishaan @hubertjb ! Just a quick question - are you sure neuralbench results are using the bugged version?
Is it possible that the results in the paper are correct but the repo had the discussed bug?

Thanks! 
Roy Segal.

guptaishaan:
Thanks, and worth separating clearly:

`config.yaml` for this task has exactly one commit in its history: `1b50f77` ("Adding NeuralBench", [#42](https://github.com/facebookresearch/neuroai/pull/42), 2026-05-06). It has been `Keystroke`-triggered since the file was created, never anything else, and `neuralbench-repo` did not exist before that commit.

* arXiv:2506.19141, the EEG Foundation Challenge 2025 paper this task's docs cite for the "0.5 to 2.5 s after stimulus onset" window, is a separate paper from a separate, code-submission competition that predates neuralbench-repo by about 11 months, with no author overlap with the NeuralBench paper below except Jean-Rémi King. Those results cannot have come from this config, so I'd expect them to be fine.
* arXiv:2605.08495, the NeuralBench paper itself, reports a reaction_time entry for Shirazi2024/HBN as part of "NeuralBench-EEG v1.0," and says its results come from NeuralBench's own pipeline rather than an external leaderboard. `docs/_static/neuralbench-results-data.json`, which backs the results table on the docs landing page, was added in the same [Adding NeuralBench #42](https://github.com/facebookresearch/neuroai/pull/42) commit and hasn't changed since. Its reaction_time entry has 16 models (8 classic + 6 foundation + 2 baseline). I've now read the typeset PDF directly rather than just the abstract/HTML rendering: the paper's own roster is actually 17 models per task — 8 task-specific + 6 foundation + 3 baselines (chance, dummy, handcrafted), confirmed by §2.9's "14 models + chance, dummy & handcrafted features baselines" (17 × 97 tasks-datasets × 3 seeds = 4,947 experiments, exactly) and by the legends on Figures 3, 4, and S2. So the JSON is short one baseline against the paper's pipeline — most likely `chance` just isn't shown on the public docs table. The 14 real models still match precisely, so that table was very likely produced by this response-locked config, just not the exact match I first thought.

The PDF doesn't state the task's stimulus/response window anywhere itself, so it can't independently confirm or rule out response-locking either way. [@hubertjb](https://github.com/hubertjb) generated that file and would know for sure whether it was ever rerun since.

So the paper you cited for the window definition should be unaffected. NeuralBench's own reported number for this task is the one I'd treat as suspect until it's rerun.

hubertjb:
> Hi!
> 
> Thanks for addressing the bug @guptaishaan @hubertjb ! Just a quick question - are you sure neuralbench results are using the bugged version? Is it possible that the results in the paper are correct but the repo had the discussed bug?
> 
> Thanks! Roy Segal.

@Roysegg I confirm this also affects the results reported in the NeuralBench paper.

@guptaishaan please let me know whether you'd like to apply the changes discussed above, or if you'd like me to pick it up. Thanks!

hubertjb:
I made the final edits and merged in #234. Thanks again @Roysegg for flagging this and @guptaishaan for the PR!

Issue #184, opened 2026-06-29
Add data loaders by study


juan-cobos:
### Description

I was looking for a simple way to load a dataset for a particular study. After exploring the codebase, I found that Data() in neuralbench provides exactly this functionality. Being able to instantiate the dataset in a single line is incredibly convenient for training and validating models.

That said, the current approach requires installing neuralbench, which seems somewhat at odds with the philosophy of neuralset as the study-focused package.

I also found the relationship between neuralset and neuralbench a bit confusing because of the circular dependency between them. I'm not sure what the best solution would be, but a couple of ideas came to mind:

- Add a simple mapping from studies to dataset configurations, making it straightforward to obtain the appropriate dataloaders.
- Alternatively, move the default dataset-loading helpers into neuralset and have neuralbench build on top of those utilities.

I'd love to hear your thoughts on the best direction. I'd also be happy to contribute an implementation if this sounds like a useful addition.

Best
Juan

### Alternatives considered

I've tried to implement the raw studies but the current device merging with task config, becomes really difficult to generalize.

teonbrooks:
hi @juan-cobos, thanks for the question.

Just make sure I cover all bases, I will walk through how to load an existing study within our catalog and how to load a new study we don't currently support.

## Loading Existing Study

If there is an existing study, we actually have a helper function within neuralset that can be use to retrieve it. Full explanation [here](https://facebookresearch.github.io/neuroai/neuralset/auto_examples/walkthrough/02_studies.html).
```python
import neuralset as ns


all_studies = ns.Study.catalog()
# list all the studies
print(all_studies.keys())

# choose a study
study = ns.Study(name="Grootswagers2022Human", path="/data_path")
```

Here, you need to know the name of the study you're interested in and where it is stored or will be stored once downloaded.

However, if you're interested in a dataset that hasn't been added to the catalog yet, check out below.

## Loading a New Study

Here's a bit of design philosophy for the neuroai monorepo. we actually have a standard way to add datasets that doesn't require installing neuralbench and it's actually through NeuralFetch. we haven't publicized the package (yet!) but it is what powers the datasets you see in both NeuralSet and NeuralBench.

https://facebookresearch.github.io/neuroai/neuralfetch/contributing.html


Let me know if this doesn't make sense or where we should update our documentation if it's not clear. We're all a bit close to the project so we would love the perspective of someone with fresh eyes to look at it with us!

juan-cobos:
[...]

PR #209, opened 2026-07-30
Add get_default_dataloaders for loading a task's data standalone

juan-cobos:
## What does this PR do?

Adds a `get_default_dataloaders` utility that returns the train/val/test
DataLoaders for a given device and task, using the same default configs as the
benchmarks:

    from neuralbench import get_default_dataloaders
    loaders = get_default_dataloaders(device="eeg", task="audiovisual_stimulus")

The idea is to train and validate models on benchmark data without inheriting
the training surroundings (model factory, Lightning module, trainer, W&B). I've
kept it minimal: it performs the same base-defaults -> task-config merge the CLI
does, then runs `Data.prepare()` on the merged `data:` subtree. Extra `data:`
keys can be passed as dotted-path overrides, e.g.
`get_default_dataloaders("eeg", "audiovisual_stimulus", batch_size=8)`.

## Tests

The last commit tests it against `eeg/audiovisual_stimulus` (Mne2013SampleEeg,
downloaded on first run). The reference builds the same loaders the long way
through the explicit neuralset API -- bare study -> `SklearnSplit` -> extractors
-> `Segmenter` -- so the two paths share no code, then asserts equal split sizes
and element-wise equal tensors on the unshuffled test split. It doubles as a
usage example.

## Related Issues

Addresses #184

## Checklist

- [x] Tests added or updated
- [ ] Docstrings updated for any changed public APIs


hubertjb:
Thanks @juan-cobos! Looks good to me overall, just one thought on the test: we try to avoid downloading or reading real studies in the neuralbench unit tests (we usually lean on the synthetic Test2024Eeg study), though that wouldn't work here since the task config points at a real study.

Since the new logic is "merge base + task config, apply overrides, build Data", it could be tested against the real YAMLs by stubbing Data.prepare, e.g.:

```python
from . import data

def test_get_default_dataloaders_merges_and_overrides(monkeypatch):
    cfg = None
    expected_loaders = {"train": object(), "val": object(), "test": object()}

    def fake_prepare(self):
        nonlocal cfg
        cfg = self
        return expected_loaders

    monkeypatch.setattr(data.Data, "prepare", fake_prepare)
    loaders = get_default_dataloaders(
        "eeg", "audiovisual_stimulus", batch_size=8, **{"neuro.frequency": 60.0}
    )
    assert loaders is expected_loaders
    assert cfg.trigger_event_type == "Stimulus"    # task config wins over base
    assert cfg.target.event_field == "description"  # task =replace= target
    assert cfg.batch_size == 8                       # base default -> kwarg override
    assert cfg.neuro.frequency == 60.0               # dotted override (base default 120.0)
```

juan-cobos:
@hubertjb Yes, you’re right! I initially just added the test to show how I’d tested it. I wasn’t planning to merge it since I wasn’t sure whether you wanted a test for the utility function.

I’ve now added your test directly to test_data.py, and it passes. I also deleted the previous test, so everything should be ready to merge! 🙂

hubertjb:
Thanks @juan-cobos , looks good to me! I'll merge for now but fyi I may make a few tweaks in a follow up PR.

Issue #63, opened 2026-05-13
Should extras conventions align across the four sibling packages?


bkowshik:
👋  Working through the docs as a new contributor and noticed three asymmetries in the `[project.optional-dependencies]` sections across the four sibling packages. Filing as a framing question — happy to send PRs to align if there's a preferred direction, or close as wontfix if the divergence is intentional.

## Current state

| Package      | Extras defined                                  |
|--------------|-------------------------------------------------|
| `neuralset`  | `tutorials`, `all`, `dev`                       |
| `neuralfetch`| `all`, `quickstart`                             |
| `neuraltrain`| `all`, `models`, `metrics`, `lightning`, `dev`  |
| `neuralbench`| `datasets`, `models`, `wandb`, `dev`, `docs`    |

### 1. `[all]` semantics differ

- `neuraltrain[all]` = `["neuraltrain[models,metrics,lightning]"]` — recursively pulls in every other extra. "All" means all.
- `neuralfetch[all]` = `["neuralset[all]", "matplotlib"]` — pulls in the sibling's `[all]` plus one library, but does NOT include `neuralfetch[quickstart]` (which holds the data-source clients: `openneuro-py`, `osfclient`, `boto3`, `pyunpack`, `praatio`, `awscli`).
- `neuralbench` has no `[all]` extra at all.

As a new user, I read `pip install neuralfetch[all]` as "all optional dependencies" and was surprised when subsequent `Study(...).download()` calls failed with `ModuleNotFoundError: pip install openneuro-py`, then `pyunpack`, then `boto3`, then `osfclient`. All of these live in `[quickstart]`, which the docs Quick start page doesn't mention installing.

### 2. `[quickstart]` is unique to `neuralfetch`

The adjacent vocabulary in the other siblings:

- `neuralset` → `[tutorials]` for runnable-example deps
- `neuralbench` → `[datasets]` for data deps
- `neuraltrain` → feature-domain names: `[models]`, `[metrics]`, `[lightning]`

The contents of `neuralfetch[quickstart]` are data-source clients (OpenNeuro, OSF, S3, archive unpacking). Would `[datasets]` or `[sources]` be more discoverable, matching `neuralbench[datasets]`?

### 3. `[dev]` is defined in three siblings but not `neuralfetch`

[...]

PR #218, opened 2026-08-04
Add NeuralFetch to the NeuralBench datasets extra

alitovchenko:
## Summary

Add `neuralfetch` to `neuralbench[datasets]`.

NeuralFetch registers the Study classes that NeuralBench benchmark configurations use.

## Problem

NeuralBench configurations reference 104 Study names. The current
`neuralbench[datasets]` installation does not install NeuralFetch. Without
NeuralFetch, 103 of those Study names are unavailable.

Issue #63 discusses the names and contents of package extras. It does not add
the missing Study provider to NeuralBench.

## Changes

- Add `neuralfetch` to the `datasets` extra.
- Show installation commands for released packages and editable monorepo checkouts.
- Explain that some studies need additional backend-specific dependencies.
- Add a regression test that checks generated distribution metadata with
  `importlib.metadata` and `packaging.Requirement`.

## Validation

- Full NeuralBench suite: 193 passed; 1 external-resource test deselected.
- Registry module: 9 passed; 1 external-resource test deselected.
- Metadata regression test: passed.
- Mypy: passed for 60 source files.
- Ruff, formatting, codespell, and `git diff --check`: passed.
- Built and installed the wheel successfully.
- Confirmed wheel metadata contains `neuralfetch; extra == "datasets"`.
- Confirmed the base installation does not add NeuralFetch.

## Compatibility

The base installation does not change. Only users who select the `datasets`
extra install NeuralFetch.

Related: #63

### DIFF
--- a/docs/neuralbench/install.md
+++ b/docs/neuralbench/install.md
@@ -41,12 +41,23 @@ pre-commit install
 
 ## Optional dependencies
 
-Extra dependency groups are available for dataset downloading and
-model loading:
+Extra dependency groups are available for dataset downloads and model loading.
+
+For a released package, run:
 
 ```bash
-pip install 'neuralbench-repo/.[datasets,models]'
+pip install 'neuralbench[datasets,models]'
 ```
+
+For an editable monorepo checkout, run:
+
+```bash
+pip install -e 'neuralbench-repo/.[datasets,models]'
+```
+
+The `datasets` extra installs NeuralFetch. NeuralFetch registers the Study
+classes that benchmark configurations use. Some studies need additional
+backend-specific dependencies.
 
 ## First-run configuration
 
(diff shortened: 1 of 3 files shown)


hubertjb:
Thanks for catching this @alitovchenko! This looks good to me overall, but a few requests before we merge - let me know if you want to handle this or would prefer I take over:

1. I'd actually move neuralfetch to the base dependencies, not the datasets extra.

2. I'd specify neuralfetch[quickstart] so we get the download dependencies too (openneuro-py, praatio, pyunpack, boto3, osfclient, awscli). This should allow dropping the datasets section entirely (synapseclient would make more sense under neuralfetch[quickstart]).

3. I'd drop the metadata test since we don't have any packaging/metadata tests elsewhere, and once neuralfetch moves to base dependencies(1 above), the extra == "datasets" check is no longer required anyway.

Also, the CLA bot flagged you'll need to sign the Meta CLA before this can be merged.

hubertjb:
Fixed and merged in #233 - thanks again @alitovchenko for flagging.

PR #225, opened 2026-08-25
fix hardcoded cluster config

nguoithichkhampha:
## What does this PR do?
Issue:
Even I set `"CLUSTER": null` in config.json but neuralbench still call to slurm when run extractor task.

This pull request updates the configuration for several task files to improve how the compute cluster is specified. Instead of hardcoding the cluster as `auto`, the configs now reference a central Python variable.

### DIFF
--- a/neuralbench-repo/neuralbench/conftest.py
+++ b/neuralbench-repo/neuralbench/conftest.py
@@ -29,9 +29,9 @@ def patch_config(monkeypatch: pytest.MonkeyPatch) -> Callable[..., None]:
         base = config_manager._default_config()
         base.update(overrides)
         monkeypatch.setattr(config_manager, "_config", base)
-        monkeypatch.setattr(config_manager, "_initialized", False)
         for key in config_manager._LAZY_CONFIG_KEYS:
             monkeypatch.delattr(config_manager, key, raising=False)
+        monkeypatch.setattr(config_manager, "_initialized", False)
 
     return _apply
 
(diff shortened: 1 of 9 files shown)


hubertjb:
Good catch and thanks for the PR @nguoithichkhampha!

Two things before we can merge:
1. `tasks/fmri/image/config.yaml` has a second `cluster: auto` (line 45, the `data.target` infra) that this PR misses, so the target feature extraction still submits to SLURM. Would you mind fixing that too?
2. We need to cover this in a unit test. `test_cli.py::test_cluster_config_wires_all_infra_clusters` covers `defaults/config.yaml` already and could be extended to loop over the configs in `TASKS` and assert no `*.infra.cluster` key remains after the merge.

Let me know if you're ok with making the changes or if we should pick it up.

nguoithichkhampha:
@hubertjb , I will update unit test to cover this fix.

nguoithichkhampha:
I also fixed an issue when run a function test inside `test_cli.py`
Screenshot 2026-09-01 at 15 59 32
the root cause from `monkeypatch.setattr(config_manager, "_initialized", False)` but then call `monkeypatch.delattr(config_manager, key, raising=False)`.
the `delattr` trigger `hasattr` then `__getattr__` -> `_ensure_initialized` and set _initialized = True

hubertjb:
Thanks @nguoithichkhampha, and good catch on the `conftest.py` ordering. We just need to fix the CI - if you can run `ruff check --fix . && ruff format .` from the repo root this should fix it.

nguoithichkhampha:
thanks @hubertjb , I have fixed the ruff lint issue.

PR #229, opened 2026-08-28
feat(neuralbench): add emg/pose regression task on EMG2Pose (NEMAR NM000281)

bruAristimunha:
## Summary

Adds `neuralbench emg pose`: the regression setting of the EMG2Pose hand-pose
benchmark (NEMAR NM000281; Salter et al., NeurIPS 2024). The task predicts
dense 20-joint hand-angle trajectories from 16-channel, 2 kHz surface EMG.

This is the paper's `regression_neuropose` setting. The autoregressive
tracking setting is intentionally out of scope because it requires model-side
state conditioning rather than a task configuration.

### Correct task configuration

- Uses 5 s EMG windows and the paper's predefined train/val/test split.
- Reads the 20 joint-angle `MISC` channels from the same BDF as the EMG and
  presents them as `(time, joints)` with `TimeMajorExtractor`.
- Corrects the BDF's `uV` header for joint channels, then converts radians to
  degrees so the reported MAE matches the paper's unit.

### Reliable EMG2Pose study loading

- Joins `split` and `generalization` from the release's required
  `sourcedata/emg2pose_metadata.csv` via each BIDS sidecar's `SourceFile`.
  This is source-only: a missing, annex-pointer, or incomplete table fails
  clearly instead of silently producing an invalid split.
- Uses `BAD_IK` BDF annotations to emit only contiguous resolved-label spans,
  and clips each recording to its `scans.tsv` duration to exclude BDF padding.
- Reuses the path resolved during BIDS discovery rather than recursively
  scanning the release again for every recording. Fully invalid recordings are
  skipped safely.

### Runnable training and public API

- Keeps a trajectory's joint axis intact when updating metrics. This lets the
  configured 20-output regression metrics log `val/mae` correctly while the
  loss continues to use its flattened representation.
- Exports and documents `TimeMajorExtractor` as part of the public
  `neuralset.extractors` API.

## Validation

- Added regression coverage for target units, trajectory metrics (including
  grouped metrics), source metadata, BIDS-path reuse, failed IK recordings,
  and `TimeMajorExtractor`.
- `22 passed`:
  `neuralbench/test_cli.py`, `neuralbench/test_pl_module.py`,
  `neuralfetch/test_salter2024emg2pose.py`, and the focused extractor test.
- `ruff@0.15.8 check` and `ruff@0.15.8 format --check` pass.
- The Sphinx HTML build completes successfully; the changed EMG task page and
  generated `TimeMajorExtractor` reference introduce no documentation
  warnings.


hubertjb:
Thanks @bruAristimunha ! Here are some curated comments from Claude's review:

```
## Blockers

2. **The configured metrics crash on trajectory shapes.** `pl_module._run_step` flattens
   3-D prediction and target to `(B, 200000)`, but the config asks for
   `get_regression_metric_configs(20)`. On torchmetrics 1.8.2, `MeanSquaredError`,
   `MeanAbsoluteError` and `PearsonCorrCoef` all raise; only `R2Score` survives. Since
   `monitor` is `val/mae`, training can't checkpoint. This needs metrics that reduce over
   time before scoring per joint.

## Correctness and design

5. **Drop the per-timeline full-tree rescan.** `_load_timeline_events` calls
   `find_matching_paths(root=bids_root, ...)` to re-find a file `iter_timelines` already
   had — a recursive walk of the whole tree, 25,253 times. Carry the resolved path in the
   timeline dict.
6. **Omit `None`-valued keys from the timeline dict.** `Study._load_one` stringifies every
   timeline value, so `None` becomes the literal `"None"`. That hits `stage`,
   `source_file`, and the BIDS entities (`run == "None"` instead of the canonical `""`),
   and puts `valid_duration` on every event row as a stringified float.
7. **Consider reading BAD_IK from `*_events.tsv`** instead of opening all 25k BDFs with
   `read_raw_bdf` during event loading, if the sidecars carry the spans.
8. **Handle the fully-failed recording.** An all-BAD_IK recording gives an empty frame,
   and `standardize_events` raises on that. Given 12.7% of frames failed IK, has this been
   ruled out?
9. **Decide on subsampling.** `emg/typing` restricts to `timeline_index < 120` for
   tractability; this config takes all 370 hours with no `query`. If deliberate, document
   it with a rough runtime.
10. **Remove the unused columns:** `moving_hand`, `held_out_user`, `held_out_stage` are
    carried but never read.

## Conventions

11. Both new `neuralfetch` files are missing the Meta copyright header.
12. The study sets neither `licence` nor `url` (30 of 33 existing studies do), so the
    CC-BY-NC-SA / UmeTrack non-commercial fact lives only in the `.rst` — even though
    `Study`'s docstring points users at those attributes.
13. `TimeMajorExtractor` is in `meta.py` as asked, but isn't exported from
    `extractors/__init__.py`, isn't in the `reference.rst` autosummary, and has no test in
    `test_meta.py`. A new general-purpose extractor wants a test more than the study
    helpers do.
14. `pose.rst` is still 88 lines and still explains split, windowing, and unit choices that
    `config.yaml` carries; the IK bullet also duplicates the `_ik_clean_spans` docstring
    nearly verbatim. The `_build_neuropose` comment still explains configuration.
15. The `.. warning::` in `pose.rst` has no blank line before it, so docutils ends the
    bullet list with an error node. CI won't catch this (no `-W`).
```

docs/neuralbench/tutorials/07_biosignal_challenge_2026/plot_overview.py
```diff
@@ -174,11 +178,10 @@
...
 #    metric the competition will use.
-# 3. **Hidden evaluation sets.** Every track is scored against a hidden
-#    test set (Alljoined Emotiv for Track 1, later Graz/BrainHero
-#    sessions for Track 2, the Muse cohort for Track 3, 100 new users
-#    for Track 4). Only the closest public training data ships here, so
-#    the numbers above are sanity checks, not leaderboard scores.
+# 3. **Hidden evaluation sets.** Tracks 1-3 are scored against hidden
+#    test sets (Alljoined Emotiv, later Graz/BrainHero sessions, and the
+#    Muse cohort). Track 4 uses EMG2Pose's published split. The numbers
+#    above are sanity checks, not leaderboard scores.
```
cedricrommel, line 180: Is this staying like this or will it be updated once we have the hidden data?

neuralbench-repo/neuralbench/tasks/emg/pose/config.yaml
```diff
@@ -0,0 +1,79 @@
...
+  =replace=: true
+  name: NeuroPose
+trainer_config:
+  monitor: val/mae
+  mode: min
+  strategy: auto
+  patience: 7
+  n_epochs: 40
+loss:
+  name: MSELoss
```
cedricrommel: The paper says it should be trained with L1Loss: c.f. "Training Setup:" paragraph in section 3.5

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
```diff
@@ -40,19 +41,6 @@
...
-    mne_bids ≥ 0.19 reads the channel types from ``channels.tsv`` and the
-    units from ``channels.tsv`` / ``_emg.json``, so the EMG channels are
-    returned in microvolts with no manual rescaling needed.
-    """
-
-    def _read(self) -> tp.Any:
-        bp = mne_bids.get_bids_path_from_fname(self.filepath)
-        return mne_bids.read_raw_bids(bp, verbose=False)
-
-
```
cedricrommel, line 55: Do we really want to remove BIDS study from emg2qwerty. Isn't it useful beyond the competition? I'm not saying we should not, just asking a question to understand and make sure this is not a mistake. emg2qwerty was in neuralbanch already.

cedricrommel (changes requested):
Nice addition of the new dataset. I just have two remarks, which might be related:
1) the emg2pose paper says their model is trained with L1 loss, while this is configuring training with MSELoss
2) for some reason you are adding the NeurPose baseline together with the emg2pose dataset. I think we want to add the vemg2pose model instead (or all three vemg2pose, NeuroPose and SensingDynamics). Maybe this explain the MSELoss mistake (I don't what loss NeuroPose reports).

docs/neuralbench/tutorials/07_biosignal_challenge_2026/plot_overview.py
bruAristimunha: This remains intentional: Track 4 uses EMG2Pose's published split, so its results are documented as sanity checks rather than hidden leaderboard scores.

neuralbench-repo/neuralbench/tasks/emg/pose/config.yaml
bruAristimunha: Fixed in 33aa5aa: the EMG2Pose task now uses L1Loss.

neuralfetch-repo/neuralfetch/studies/sivakumar2024emg2qwerty.py
bruAristimunha: It remains available: the study-local BidsEmg subclass was replaced with the reusable neuralfetch.bids.BidsEmg implementation.

bruAristimunha:
Addressed Cedric’s model review in 33aa5aa: EMG-to-Pose now uses the paper’s VEMG2Pose regression baseline and L1 loss; the EMG model registry and tutorial were updated accordingly. The branch also incorporates current `main` without a force-push.

neuralfetch-repo/neuralfetch/studies/salter2024emg2pose.py
```diff
@@ -0,0 +1,193 @@
...
+            if onset < valid and onset + duration > 0
+        )
+        spans, cursor = [], 0.0
+        for start, stop in bad:
+            if cursor < start:
+                spans.append((cursor, start - cursor))
+            cursor = max(cursor, stop)
+        if cursor < valid:
+            spans.append((cursor, valid - cursor))
+        return spans
```
bruAristimunha: clarify this... not super linear..

bruAristimunha: replace with not valid annotation, and mask any invalid in the loss and in the prediction

neuralfetch-repo/neuralfetch/studies/salter2024emg2pose.py
bruAristimunha, line 1: Special loader can solve....

bruAristimunha: Replace the EMG bids for this... 

Read the data, load the target, override the target with -1000, and then the data cache on disk will have this.

bruAristimunha (commented):
increase braindecode version

neuralfetch-repo/neuralfetch/studies/salter2024emg2pose.py
bruAristimunha: Done in e90363e: BAD_IK target samples are stored as -1000 in the cache, and the shared BrainModule masks them from sequence loss, metrics, and saved predictions/targets.

neuralfetch-repo/neuralfetch/studies/salter2024emg2pose.py
bruAristimunha: Done in e90363e: SpecialLoader now reads the BIDS annotations, overwrites BAD_IK target spans with -1000 before writing the cache, crops the padded tail, and keeps one continuous BidsEmg event.

bruAristimunha:
Updated the NeuralBench dependency to `braindecode>=1.8.1`. The published PyPI wheel now includes VEMG2Pose, NeuroPose, and SensingDynamics; all three instantiate and run from the release package.

neuralfetch-repo/neuralfetch/studies/salter2024emg2pose.py
```diff
@@ -0,0 +1,254 @@
...
+        for annotation in raw.annotations:
+            if annotation["description"] != self.IK_ANNOTATION:
+                continue
+            start, stop = raw.time_as_index(
+                [annotation["onset"], annotation["onset"] + annotation["duration"]],
+                use_rounding=True,
+            )
+            invalid[max(0, start) : min(raw.n_times, stop)] = True
+        if invalid.any():
+            raw.apply_function(lambda data: np.where(invalid, np.nan, data), picks="misc")
```
cedricrommel, line 253: Nice progress on BAD_IK masking.
One remaining parity gap with public emg2pose: upstream defaults to skip_ik_failures=True semantics (windows built from contiguous IK-valid spans), not only downstream loss/metric masking.
To reproduce the public repo behavior, this should be the default here too. We should add an option to fully drop windows with IK.

neuralbench-repo/neuralbench/pl_module.py
```diff
@@ -160,7 +160,34 @@ def _run_step(
...
+            # Models that consume left context emit only the window's valid tail.
+            y_true = y_true[:, -y_pred.shape[1] :]
+            # NaN marks a frame the study could not label (e.g. unresolved
+            # inverse kinematics); it must reach neither the loss nor a metric.
+            valid = ~y_true.isnan().any(dim=-1)
+            metric_pred = loss_pred = y_pred[valid]
+            metric_true = loss_true = y_true[valid]
+            # Metric rows are frames, not windows: (B, 1) -> (B, T).
+            metric_subjects = metric_subjects.reshape(-1, 1).expand_as(valid)[valid]
+            y_pred = y_pred.masked_fill(~valid.unsqueeze(-1), torch.nan)
```
cedricrommel, line 197: Just to clarify: this is only part of the IK failures handling mechanism implemented in the official emg2pose repo. We should keep it, but as I mentioned in my other comment, we also need to optionally drop windows presenting any IK failure (with the default set to True, i.e., dropping).

cedricrommel (changes requested):
Very nice reproduction. I think we are almost there! Thanks for pushing this through!!

neuralfetch-repo/neuralfetch/studies/salter2024emg2pose.py
hubertjb: Thanks @cedricrommel! I added dataset-level filtering of the segments with invalid IK values, so we have both on by default.

hubertjb:
Thanks @bruAristimunha and @cedricrommel ! Merging this first version. @bruAristimunha is running some validation runs, we can revisit if need be in a second PR.

PR #237, opened 2026-09-04
EEG/EMG Foundation Challenge 2026 starter kit: MAE pretraining example + training page

hubertjb:
Starter kit for the [EEG/EMG Foundation Challenge 2026](https://neural-interfaces26.github.io/): a self-supervised pretraining path that runs end to end, from unlabelled EEG to a NeuralBench score, plus the challenge pages that frame it.

## Code

**`MaeEncoder` (`neuraltrain/models/mae.py`)** — masked prediction over channel-time patches, one token per channel-patch. Channels are identified by a 3D position embedding rather than by index, and channels with no valid position (the zero-padding that montages with fewer channels get) are dropped from the attention, so a single encoder pretrains across the challenge's 63-, 60-, 19- and 2-channel studies. Encoder-only by design: the decoder is left to participants, and the module scores the hidden patches with a plain `MSELoss`. `channel_positions` has to describe the channels of the input, and is checked against them rather than trusted.

**`neuraltrain-repo/ssl_example/`** — the self-supervised counterpart to `project_example`, same `main.py` / `grids/` / `README.md` shape, with pretraining driven by `mae_module.py`. Three things differ, and they are what self-supervision changes: windows tile the recording on a stride rather than following events, the segmenter has an `"input"` extractor and no `"target"` one, and whole subjects are held out so validation measures generalisation to an unseen recording. Defaults pretrain on the studies of tracks 1-3 plus a resting-state study, preprocessed exactly as `neuralbench` does, multi-GPU under DDP, with optional W&B.

**`neuralbench/models/mae.yaml`** — rebuilds the encoder headless for `-m mae --checkpoint <path>` and pins the `dim` and `patch_size` a checkpoint has to match; channel count need not match, since channels are read from their positions. How the encoder is then adapted is left to `-w`, as for every other model: `-w linear_probe_mean` freezes it and trains a linear probe on the mean-pooled tokens, `-w lora_r4_flatten` uses the `lora_target_modules` the YAML names, and no `-w` fine-tunes. A `dim` or `patch_size` mismatch warns and keeps the randomly initialised layer.

## Docs

A new training page (`plot_pretrain_mae.py`) covers caching the data, pretraining, probing the result, and the `evaluate_model` route for a model that lives in the participant's own script. The existing challenge pages now follow the competition website for names, track titles, shifts, and headline metrics, and carry no dates or NeurIPS framing, since the competition runs as part of the Brain and Body Foundation Model workshop. `install.md` drops the removed `models` extra and says W&B is opt-in.

## Infra

- `neuralbench` depends on `braindecode[hub]`: `braindecode` stubs out `from_pretrained` unless `huggingface_hub` is installed, and every pretrained foundation model goes through it. That made the `[models]` extra redundant, so it is gone.
- `neuralbench` depends on `x_transformers`: `MaeEncoder` always builds `neuraltrain`'s `TransformerEncoder`, which reaches `neuralbench` only through `neuraltrain[models]`, declared by nothing. Without it a base install advertises `-m mae` and cannot build it — the same hole `braindecode[hub]` closes for the pretrained models.
- Both `neuraltrain` examples are packages, so bare `pytest` in `neuraltrain-repo` collects them. `from ..main import Experiment` had no package to resolve against, which already broke collection for `project_example` on `main`. Their entry points are named `run`, since pytest collected `test_run` as a test and it takes an argument.
- CI gains a `neuraltrain ssl example` step, running the unit tests and the end-to-end debug run, since examples live outside the package and `pytest neuraltrain` misses them.
- Drive-by fix: the swav image-similarity metric loads with `torch.hub.load(..., skip_validation=True)`. `torch.hub` checks that the ref is not from a fork by calling the GitHub API unauthenticated, which rate-limits by IP and was failing CI with a 403; the repo and ref here are fixed rather than input, so the check has nothing to catch.

## Verified

- A fresh conda env following only the published docs: pretraining saves `encoder.ckpt`, and `neuralbench eeg dementia_diagnosis -m mae --checkpoint <path> -w linear_probe_mean` scores it. Reloading the checkpoint into the encoder reports zero missing and zero unexpected keys.
- Full pretraining across the four studies on 8 GPUs under DDP, with W&B logging; the loss follows the `OneCycleLR` schedule, dropping through the first epoch and again as the LR anneals.
- `-w linear_probe_mean` freezes all 3,225,609 encoder parameters now that `mae.yaml` no longer pins `layers_to_unfreeze`, and the four `lora_target_modules` resolve to real submodules.
- `ruff`, `mypy`, the MAE tests, the `ssl_example` debug run, and a docs build pass. Bare `pytest project_example ssl_example` collects cleanly, where it used to error.

## Notes for review

- `mae.yaml` adds an entry to the shared model registry; happy to move it behind an entry point if that registry should stay curated.
- Making the examples packages touches `project_example`, whose collection was already broken on `main`. Happy to split that into its own PR if it does not belong here.

## Open

The official track corpora are released through NeuralBench when submissions open, so the track pages still run on the closest open datasets. `plot_overview.py` lists what is missing.


neuralbench-repo/pyproject.toml
jraugel:
```suggestion
  "neuraltrain",
  # MaeEncoder builds its transformer through neuraltrain's x_transformers wrapper
  "x_transformers>=1.27.20",
```

`x_transformers` seems to only ship in `neuraltrain[models]`, which `neuralbench` doesn't request,  so `-m mae` can't build on a base install. Maybe it's worth naming the one package rather than the extra (which drags in unused models like `dalle2_pytorch`/`vit_pytorch`)

PR #253, opened 2026-09-10
fix: support Conformer padding lengths

jarod1212:
Allow callers to provide sequence lengths to the Conformer wrapper while preserving the existing full-length default. This enables correct attention masking for padded batches without changing existing callers.

### DIFF
--- a/neuraltrain-repo/neuraltrain/models/test_conformer.py
+++ b/neuraltrain-repo/neuraltrain/models/test_conformer.py
@@ -45,3 +45,28 @@ def test_conformer(fake_input, use_default_config):
 
     # Check shape of the output
     assert out.shape == (batch_size, n_times, n_in_channels)
+
+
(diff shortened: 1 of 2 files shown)


neuraltrain-repo/neuraltrain/models/test_conformer.py
```diff
@@ -45,3 +45,28 @@ def test_conformer(fake_input, use_default_config):
 
     # Check shape of the output
     assert out.shape == (batch_size, n_times, n_in_channels)
+
+
+def test_conformer_uses_lengths():
```
jrapin: can this tests be fused into another consolidated one?
it may not be, but I feel it is overly long for such a small update, and tests have a maintenance cost

jarod1212:
I think it's better now. Ready to merge. @jrapin

jrapin (approved):

PR #258, opened 2026-09-11
neuralfetch: unify selective download (include/exclude), fold in #25, fix brainai#1751

### LINKED ISSUE #51: [neuralfetch] Refactor selective folder/file downloading
> we currently two ways to download from OpenNeuro, (1) openneuro-py to download, which uses pooch underneath the hood and (2) datalad. I think most of our new studies are using (1). 
> 
> 
> 
> I think for now the paths approach is fine, but may be refactored later. 

 _Originally posted by @teonbrooks in [#25](https://github.com/facebookresearch/neuroai/pull/25/changes#r3168622852)_

Currently we have divergent ways to filter selective files and folders when downloading. We should unify the approach for consistency.

### LINKED ISSUE #257: [neuralfetch] Proposal: unify selective download behind include/exclude globs on BaseDownload
Concrete proposal following up on the upstream neuroai issue
[facebookresearch/neuroai#51](https://github.com/facebookresearch/neuroai/issues/51).

**Status.** The two-tier `overwrite` standardization this work was sequenced
behind has now merged (neuroai [#222](https://github.com/facebookresearch/neuroai/pull/222) / [#224](https://github.com/facebookresearch/neuroai/pull/224), brainai [#3241](https://github.com/fairinternal/brainai/pull/3241)), and the neuralfetch CLI has landed ([#223](https://github.com/facebookresearch/neuroai/pull/223)).

Selective download is therefore unblocked and is the clean next step on the same
`BaseDownload` backends. The `_download(self, overwrite: bool = False)` signature
in the snippet below reflects the merged state.

## The divergence today

Every backend in `neuralfetch/download.py` invents its own selection vocabulary:

- `Openneuro` — `include: list[str] | None` (openneuro-py glob semantics), plus `excluded_patterns: list[str]` that is **declared on the model but never read** by `_download`. Any caller passing it gets a silent no-op and downloads the entire dataset. `openneuro.download()` natively supports `exclude=`, so the capability is there, just unwired.
- `Datalad` — `folders: list[str | Wildcard]`, where a bare `str` is a literal folder and `Wildcard(folder=...)` is a glob.
- `S3` / `Physionet` — `prefix`, `files_with_destinations: list[tuple[str, str]]`, `skip_existing`.
- `Osf` — `storage_inds: list[int]` (selects storages, not paths).
- `Globus` — selection only by making `study` the remote directory root, then walking it recursively.
- `Huggingface` — nothing exposed, though `snapshot_download` supports `allow_patterns`/`ignore_patterns`.
- `Dandi`, `Zenodo`, `Figshare`, `Dryad`, `Synapse`, `Eegdash`, `Donders` — no selection at all, all-or-nothing.

The silent no-op, in full:

```python
class Openneuro(BaseDownload):
    excluded_patterns: list[str] = []      # declared, still never read
    include: list[str] | None = None       # the only selector actually used
    nworkers: int = 5

    def _download(self, overwrite: bool = False) -> None:
        import openneuro as on

        if not overwrite and self._has_local_data():
            return

        on.download(
            dataset=self.study,
            target_dir=self._dl_dir,
            include=self.include,          # excluded_patterns never reaches here
            max_concurrent_downloads=self.nworkers,
        )
```

Practical consequences: a study can believe it filtered and still pull hundreds of GB; a filter cannot survive switching a study from one backend to another; and exclusion is inexpressible for most backends, so studies either over-download or hand-roll bespoke logic.

## Proposal

Two fields on `BaseDownload`, inherited by every backend:

```python
class BaseDownload(_Module):
    include: list[str] = []   # dataset-relative POSIX globs; empty means everything
    exclude: list[str] = []   # applied after include; exclude wins
```

Matching is centralized in one helper so semantics are identical everywhere: patterns match dataset-relative POSIX paths, `*` / `?` / `**` supported (aligned with openneuro-py's documented semantics so existing patterns stay valid), empty `include` means everything, `exclude` beats `include`.

Per-backend mapping:

- Native pass-through — `Openneuro` to `include`/`exclude`; `Huggingface` to `allow_patterns`/`ignore_patterns`.
- Filter the job list they already build (a few lines each) — `S3._resolve_prefix` / `_resolve_mapped`, `Globus._resolve_walk`, `Osf._download`.
- Filter the API file listing — `Zenodo`, `Figshare`, `Dryad`.
- Translate to native selection — `Datalad` include globs to `datalad get` patterns, keeping `folders`/`Wildcard` working but marked superseded.
- No usable native filter (`Dandi`, `Synapse`, `Eegdash`, `Donders`) — raise `NotImplementedError` when `include`/`exclude` is set, instead of silently ignoring it, which is exactly the `excluded_patterns` failure mode.

Existing fields that serve a different purpose stay: `prefix` (remote root), `files_with_destinations` (key remapping), `storage_inds` (OSF storages).

## Feasibility

High, and the blast radius is small. 9 of 13 backends can support the interface, and 8 of those already materialize a candidate file list before transferring, so most of the work is a shared matcher plus a handful of call sites. Current usage in `neuralfetch/studies`: 5 `include=` call sites and 1 `folders=`/`Wildcard` site.

Migration: replace `excluded_patterns` with `exclude`, convert those 6 call sites, add matcher unit tests plus one filter test per backend family.

## Open questions

- Should `**` follow openneuro-py/fnmatch semantics or full `PurePath` glob semantics?
- Should patterns be validated against a dry-run listing so a typo warns ("pattern matched 0 files") instead of silently downloading nothing?

teonbrooks:
## Summary

Unifies **selective downloading** across every neuralfetch download backend behind two inherited glob fields on `BaseDownload` — `include` and `exclude` (dataset-relative POSIX globs; empty `include` means "everything", `exclude` is applied after and wins) — resolved through a single centralized matcher, `_selects()`. This completes the proposal in #257 (follow-up to the now-closed #51).

It also folds in and supersedes #25 (thanks @bpinsard) — his commits are **cherry-picked here with authorship preserved** — moving Datalad off `subprocess` onto the `datalad` Python API. As a natural consequence, a missing `datalad` install now **errors instead of silently warning and writing a bogus success marker over an empty folder** (fixes fairinternal/brainai#1751).

## What changed

- **`BaseDownload`**: new `include`/`exclude` fields, `_globs_match()` helper (openneuro-py glob semantics: `*`/`?`/`**`, folder patterns select their subtree), `_selects()` matcher, and `_reject_selection_filters()`.
- **Datalad / Gin**: fetch selected paths via `datalad.api` (built on @bpinsard's rewrite). Clone still lands in `download/<repo_name>/` (preserving existing study layout). `datalad` added to `requirements`. Gin keeps its git-annex web-URL registration, now driven by `include`/`exclude`.
- **Native pass-through**: `Openneuro` (`include`/`exclude`), `Huggingface` (`allow_patterns`/`ignore_patterns`). Removes the dead `Openneuro.excluded_patterns` field.
- **List/API filtering** through `_selects`: `S3`/`Physionet`, `Globus`, `Osf`, `Zenodo`, `Figshare`, `Dryad`.
- **No native filter**: `Dandi`, `Synapse`, `Eegdash`, `Donders` now raise `NotImplementedError` when `include`/`exclude` is set, instead of silently ignoring it.
- **Call sites**: `hebart2023things` migrated from `folders=[Wildcard("sub-*")]` to `include=["sub-*"]`; the existing `include=` study call sites are unchanged.

## Tests

- Matcher unit tests (`_globs_match` / `_selects` precedence).
- Per-backend selection wiring (`S3._resolve_prefix`, `Datalad._selected_paths`, `Gin` end-to-end).
- `NotImplementedError` for the unsupported backends.
- Regression: `download()` never writes a success marker when `_download` raises, and a missing `datalad` install leaves the folder empty (neuroai#51 / brainai#1751).

All of `test_download.py` passes; `ruff`, `ruff-format`, and `codespell` are clean.

## Issues

- Closes #257
- Supersedes #25 (its commits are included with credit to @bpinsard — this PR can replace it)
- Fixes fairinternal/brainai#1751 — that issue lives in another repo and must be closed manually after this syncs into brainai.

### DIFF
--- a/neuralfetch-repo/neuralfetch/download.py
+++ b/neuralfetch-repo/neuralfetch/download.py
@@ -416,10 +416,6 @@ class BaseDownload(_Module):
         raise NotImplementedError
 
 
-class Wildcard(BaseModel):
-    folder: str
-
-
 class S3(BaseDownload):
     """Download files from an S3 bucket via ``boto3``.
 
@@ -590,8 +586,8 @@ class Datalad(BaseDownload):
         URL of the DataLad repository to clone.
     threads : int
         Number of parallel download threads for ``datalad get`` (default: 1).
-    folders : list of str or Wildcard
-        Directories or glob patterns to retrieve after cloning.  When empty,
+    paths : list of str
+        paths or glob patterns to retrieve after cloning.  When empty,
         all content is retrieved via ``datalad get "*"``.
     """
 
@@ -600,8 +596,8 @@ class Datalad(BaseDownload):
     repo_url: str
     # number of threads used for the datalad operations
     threads: int = 1
-    # list of folders (or wildcards) to actually clone
-    folders: list[str | Wildcard] = []
+    # list of path to actually fetch
+    paths: list[str] = []
 
     @pydantic.computed_field  # type: ignore
     @property
@@ -611,43 +607,6 @@ class Datalad(BaseDownload):
         if Path(repo_name).suffix == ".git":
             repo_name = repo_name[:-4]
         return repo_name
-
-    def _datalad(self, cmd: str, path: Path | str) -> None:
-        # TODO: do not use subprocess
-        proc = subprocess.run(
-            cmd, cwd=str(path), capture_output=True, text=True, shell=True
-        )
-        if "install(error)" in proc.stdout:
-            logging.warning("Potential error in datalad clone:\n> %s", proc.stdout)
-        if proc.stderr:
-            # NOTE: stderr might be populated even in success case
-            logging.warning("Potential error in datalad clone:\n> %s", proc.stderr)
-            # raise RuntimeError(f"Clone Failed: {proc.stderr}")
-
-    def _dl_item(self, cur_path: Path | str) -> None:
-        threads_ = f" -J {self.threads}" if self.threads > 1 else ""
-        cmd = f'datalad get "{cur_path}"{threads_}'
-        self._datalad(cmd, self._dl_dir / self.repo_name)
-
-    def _expand_folders(self) -> list[Path]:
-        """Resolve :attr:`folders` into absolute paths inside the cloned repo.
-
-        Wildcard entries are glob-expanded against the working tree (which
-        already contains the BIDS layout because ``datalad clone`` populates
-        directory structure and small files even on annex-ignored remotes).
-        Bare strings are appended as-is.
-        """
-        folders = self.folders if self.folders else [Wildcard(folder="*")]
-        all_folders: list[Path] = []
-        for folder in folders:
-            if isinstance(folder, Wildcard):
-                all_folders += [
-                    Path(str(p))
-                    for p in glob(str(self._dl_dir / self.repo_name / folder.folder))
-                ]
-            else:
-                all_folders += [self._dl_dir / self.repo_name / folder]  # type: ignore[list-item]
-        return all_folders
 
     def _download(self, overwrite: bool = False) -> None:
         """Downloads data from datalab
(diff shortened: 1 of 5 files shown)


teonbrooks:
thanks @bpinsard for your PR! I've folded it into this one with an update to the BaseDownload class to support include/exclude parameters.

PR #264, opened 2026-09-15
neuralfetch(bug): keep Ghassemi2018You annotations on the record's time axis

### LINKED ISSUE #262: neuralfetch: Ghassemi2018You sleep-stage annotations are shifted earlier by the unscored lead-in
### Description

`Ghassemi2018You._load_sleep_events` subtracts the onset of the first annotation from every annotation (the comment says it assumes annotations start at time 0). In the Challenge 2018 records scoring starts minutes into the recording, so every sleep stage, and the first-N2 onset that `AddSleepOnsetTargets` derives from it, is shifted earlier by a record-specific amount.

Examples (first N2 onset in seconds from the start of the record, WFDB sample index / 200 Hz vs what the loader returns): tr03-0005 780 vs 420; tr03-0029 2100 vs 1560; tr05-0028 2310 vs 1920.

Effect on the sleep_onset task with the 994 labelled training records: a crop-only gradient-boosting model on spectral features scores bMAE 160.4 against 160.1 for a constant predictor, i.e. no signal; with the annotations kept at their absolute sample index the same model scores 143.0 and a sequence model 119.8 (5-fold person-grouped CV, 3 repeats). If the published Track 3 baselines on ghassemi2018you used this loader, they were trained and scored on shifted labels.

Expected: annotation onsets equal to the WFDB sample index divided by the sampling rate, with no subtraction of the first onset. 

### Steps to reproduce

```python
neuralfetch 0.3.1, record tr03-0005 from physionet-open/challenge-2018/1.0.0/training/

import wfdb
ann = wfdb.rdann("tr03-0005/tr03-0005", "arousal")
# first stage annotation at sample 72000 (360 s); first N2 at sample 156000 (780 s)

from neuralfetch.studies.ghassemi2018you import Ghassemi2018You
study = Ghassemi2018You()
timeline = next(t for t in study.iter_timelines() if "tr03-0005" in str(t))
events = study._load_sleep_events(timeline)
events[events.stage == "N2"].start.iloc[0]   # 420.0, expected 780.0
```

### Environment

- OS:
- Python version:
- Relevant package versions:

terezaokalova:
## What does this PR do?

`Ghassemi2018You._load_sleep_events` subtracted the onset of the first annotation from every annotation, assuming annotations start at time 0. Challenge 2018 scoring starts minutes into each record, so every stage and arousal, and the first N2 that `AddSleepOnsetTargets` derives, was shifted earlier by a record-specific lead-in (tr03-0005: N2 at 420 s instead of 780 s). This keeps the WFDB sample index / fs as the start time and adds a regression test on a synthetic `.arousal` file whose first annotation is at 360 s; the test fails on `main` and passes here.

## Related Issues

Fixes #262 (reported by myself)

## Checklist

- [x] Tests added or updated
- [ ] Docstrings updated for any changed public APIs (no public API changed)
- [x] `pytest` and `mypy` pass locally (`pytest neuralfetch`: 97 passed; `mypy` shows only the pre-existing `neuralset.*` import-not-found errors of a non-strict editable install, same as on `main`)


teonbrooks:
Hi @terezaokalova, thanks for creating the issue and the PR. this looks good, good find!

I will re-run this tomorrow on some internal tests to evaluate and report back.

hubertjb:
Thanks a lot for reporting @terezaokalova! I re-ran the different sleep tasks without/with your fix and can confirm the bug and the fix.

On all three tasks, the performance improves reliably vs. chance/dummy with the fix:

**sleep_stage** (balanced accuracy %, higher better)

| | chance | dummy | EEGNet | LUNA |
|---|---|---|---|---|
| before | 19.8 | 20.0 | 47.2 | 50.6 |
| after | 19.6 | 20.0 | **73.2** | **78.8** |

**sleep_arousal** (balanced accuracy %, higher better)

| | chance | dummy | EEGNet | LUNA |
|---|---|---|---|---|
| before | 49.9 | 50.0 | 55.6 | 59.8 |
| after | 50.0 | 50.0 | **61.1** | **68.7** |

**sleep_onset** (balanced MAE in seconds, lower better)

| | chance | dummy | EEGNet | LUNA |
|---|---|---|---|---|
| before | 200.4 | 263.3 | 161.9 | 162.0 |
| after | 203.4 | 287.4 | **146.0** | **136.9** |

Otherwise I'd only have two nitpicks on the PR: 

1. The conversion from `mne.Annotations` to pd.DataFrame could be dropped, and `mne_annots.onset` / `.duration` / `.description` could be read directly to avoid setting `time_format`, e.g.
```python
annots_df = pd.DataFrame(
    {
        "start": mne_annots.onset,
        "duration": mne_annots.duration,
        "description": mne_annots.description,
    }
)
```

2. Let's bump up the study version since this change will invalidate the study cache:
```python
    def model_post_init(self, log__: tp.Any) -> None:
        # v2: annotation onsets are no longer rebased on the first annotation.
        self.version = "v2"
```

Please let me know if you'd like to make the changes or if I should pick it up.

hubertjb:
Thanks again @terezaokalova! I implemented the two small changes before merging.

PR #269, opened 2026-09-17
neuralset(events): add check_event_durations unit-mismatch guard

teonbrooks:
## Why
BIDS `_events.tsv` `duration` is seconds, but some datasets store sample counts or ms there (ds004357: ms; ds004579: samples). A range check can't see it — values still look small.

## What
Adds `check_event_durations(df, context=None)`: two heuristics — durations that overlap the next same-type event by a large ratio, and a single whole-number duration pasted across irregularly-spaced events. `standardize_events` calls it as a backstop; callers reading a known file pass `context` to name it in the warning.


teonbrooks:
closing this PR for now

bruAristimunha:
hey @teonbrooks, we can fix this at mne-bids if you want

bruAristimunha:
we can fix at dataset level too with nemar datasets

teonbrooks:
@bruAristimunha, good idea, we should probably fix it upstream in mne-bids.

PR #275, opened 2026-09-20
FIX slow unzip on password protected archive

tomMoral:
Password-protected zips (e.g. THINGS-images) decrypt via CPython's pure-Python ZipCrypto, which is single-threaded and CPU-bound.
Prefer a native C `7z`/`unzip` when on PATH, else fan members out over processes; plain zips are unchanged.
Measured ~0.4 MB/s/core in pure Python (~515 min for 12 GB) → ~6 min on ~90 cores (native is faster still).


kingjr:
Multiprocessing fallback relies on zipfile which may be challenging to AES-encrypted ZIPs. 
Else LGTM

tomMoral:
Currently, it relies on zipfile I think so I would say this is an orthogonal matter.

